PR #27, opened 2018-12-27
support get forward address from configuration_query_by_index_response

qinzuoyan:
related pr https://github.com/XiaoMi/rdsn/pull/212

with this feature, we can extend pegasus meta server nodes.

if meta_list in client config is: 127.0.0.1:34601,
but pegasus cluser's meta list is: 127.0.0.1:34601,127.0.0.1:34602, and 127.0.0.1:34602 is the leader,
it also can work.


src/main/java/com/xiaomi/infra/pegasus/rpc/async/MetaSession.java
```diff
@@ -50,6 +51,13 @@ static public final error_types getMetaServiceError(client_operator metaQueryOp)
         return op.get_response().getErr().errno;
     }
 
+    static public final rpc_address getMetaServiceFrowardAddress(client_operator metaQueryOp) {
```
neverchanje: 拼错：`getMetaServiceForwardAddress`

PR #62, opened 2019-11-01
feat: add more exception info on synchronous APIs

neverchanje:
In most synchronous APIs of PegasusTable, when `Future.await` throws `InterruptedException` and `TimeoutException`, they are wrapped under PException with simply coded ERR_TIMEOUT. This makes it hard to distinguish the real problem.

So here we add more details on each exception, InterruptedException is coded ERR_THREAD_INTERRUPTED, and information like "tableName" and "timeout ms" is complemented.


Shuo-Jia:
这个pr就不加版本信息了？

neverchanje:
> 这个pr就不加版本信息了是吗？

下个 PR 再考虑吧

neverchanje:
下个PR加吧，版本很重要，在新的PR加比较好

PR #63, opened 2019-11-04
feat: have package version prefixed in PException

neverchanje:
Package version is necessary for debugging our clients in Xiaomi, where hundreds of users are using pegasus-java-client in different versions. Some may misuse but insist they are using the latest version (usually with less bugs). With version included in `PException`, we can save time for digging down to the root cause.

## How does it work

We use **maven-jar-plugin** to inject the version during packaging. To obtain the version we use
```java
PException.class.getPackage().getImplementationVersion()
```
When unit test the version is set "{version}" as a place-holder.

Finally the exception message of a PException is like:

```
com.xiaomi.infra.pegasus.client.PException: 1.11.8-thrift-0.11.0-inlined: com.xiaomi.infra.pegasus.rpc.ReplicationException: ERR_OBJECT_NOT_FOUND: [table=temp,operation=put,replicaServer=127.0.0.1:34801,gpid=(gpid(2.1))] The replica server doesn't serve this partition!"
````

### DIFF
--- a/src/test/java/com/xiaomi/infra/pegasus/client/TestPException.java
+++ b/src/test/java/com/xiaomi/infra/pegasus/client/TestPException.java
@@ -13,7 +13,7 @@ public class TestPException {
   public void testThreadInterrupted() throws Exception {
     PException ex = PException.threadInterrupted("test", new InterruptedException("intxxx"));
     Assert.assertEquals(
-        "com.xiaomi.infra.pegasus.rpc.ReplicationException: ERR_THREAD_INTERRUPTED: [table=test] Thread was interrupted: intxxx",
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/test/java/com/xiaomi/infra/pegasus/client/TestPException.java [hunk 1]
```diff
@@ -13,15 +23,55 @@
...
+
+  @Test
+  public void testVersion() {
+    PException ex = new PException("test");
+    Assert.assertEquals("{version}: test", ex.getMessage());
+
+    ex = new PException("test", new TimeoutException());
+    Assert.assertEquals("{version}: test", ex.getMessage());
+  }
+
```
Shuo-Jia: 这个测试和之前两个测试其实已经重了，需要单独再测试一下吗？

neverchanje: 不重，这也是为了测试 PException 的两个构造函数，已加注释

PR #65, opened 2019-11-07
fix: fix bug where timeout in exception info is 0

Shuo-Jia:


src/test/java/com/xiaomi/infra/pegasus/client/TestPException.java
```diff
@@ -76,4 +78,32 @@ public void testHandleReplicationException() throws Exception {
     }
     Assert.fail();
   }
+
+  @Test
+  public void testTimeOutIsZero() throws Exception {
```
neverchanje: 加一行注释：

```
// ensure "PException ERR_TIMEOUT" is thrown with the real timeout value, when user given timeout is 0.
```

PR #82, opened 2019-12-30
refactor(rpc): refactor request meta to adapt to new version of rpc protocol

levy5307:
Refactor our thrift message parser by redesigning the request header format, to support adding is_backup_request new field.

**Side effect**:
This pull request will import incompatibility.
> old/new client --> new server ok.
old client --> old server is ok, too.
but new client --> old server is not ok.
Because the connection will be disconnected by server. So we can`t get any information about version incompatibility, it simply disconnects the connection.


idl/base.thrift
```diff
@@ -22,3 +22,10 @@ struct rpc_address
 struct gpid
 {
 }
+
+struct request_meta {
```
hycdong: Why add `request_meta` in `base.thrift` rather than `replication.thrift`?

src/main/java/com/xiaomi/infra/pegasus/rpc/ThriftHeader.java
```diff
@@ -6,32 +6,18 @@
 import java.nio.ByteBuffer;
 
 public class ThriftHeader {
-  public static final int HEADER_LENGTH = 48;
+  public static final int HEADER_LENGTH = 16;
```
hycdong: Will update `HEADER_LENGTH` directly leading to incompatibility?

src/main/java/com/xiaomi/infra/pegasus/rpc/ThriftHeader.java
```diff
@@ -6,32 +6,18 @@
...
   static final byte[] HEADER_TYPE = {'T', 'H', 'F', 'T'};
-  public int hdr_version = 0;
```
hycdong: Is it safety to remove varieties such as `hdr_version`, `header_crc32`,  `body_crc32 `, `thread_hash` directly?

scripts/travis.sh
```diff
@@ -26,8 +26,8 @@ if [[ $(git status -s) ]]; then
     exit 1
 fi
 
-PEGASUS_PKG="pegasus-tools-1.11.6-9f4e5ae-glibc2.12-release"
-PEGASUS_PKG_URL="https://github.com/XiaoMi/pegasus/releases/download/v1.11.6/pegasus-tools-1.11.6-9f4e5ae-glibc2.12-release.tar.gz"
+PEGASUS_PKG="pegasus-tools-1.13.SNAPSHOT-695b366-glibc2.17-release"
```
hycdong: I suggest adding some comments here to explain why update pegasus-tools to a SNAPSHOT version, and the time when the tools can be updated into a release version.

src/main/java/com/xiaomi/infra/pegasus/rpc/ThriftHeader.java
levy5307: What does imcompatibility mean? 
The new version client can access old version server. I have already tested it.

src/main/java/com/xiaomi/infra/pegasus/rpc/ThriftHeader.java
levy5307: same as above

src/main/java/com/xiaomi/infra/pegasus/rpc/ThriftHeader.java
hycdong: Okay~ Your comment has already answered my question~ I have another one. What about the old version client try to access the new version server?

levy5307: Sorry I have made a mistake.
old/new client --> new client is ok.
old client --> old client is ok, too.
but new client --> old server is not ok, because the connection will be disconnected by server. So we can`t get any information about version imcompatibility, It simply disconnects the connection.

idl/replication.thrift
```diff
@@ -31,3 +31,10 @@ struct query_cfg_response
     4:bool is_stateful;
     5:list<partition_configuration> partitions;
 }
+
+struct request_meta {
```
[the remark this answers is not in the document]
levy5307: I don`t suggest to use optional in java client. We use optional in pegasus server, because we want it to have good compatibility, so we can delete some data in later java client versions. But in this version of java client, these data is a must.

src/main/java/com/xiaomi/infra/pegasus/operator/client_operator.java
```diff
@@ -5,33 +5,39 @@
...
-import com.xiaomi.infra.pegasus.tools.Tools;
 
 public abstract class client_operator {
   public client_operator(gpid gpid, String tableName) {
     this.header = new ThriftHeader();
-    this.header.app_id = gpid.get_app_id();
-    this.header.partition_index = gpid.get_pidx();
+    this.meta = new request_meta();
+    this.meta.setApp_id(gpid.get_app_id());
+    this.meta.setPartition_index(gpid.get_pidx());
```
Shuo-Jia: `setApp_id` and `setPartition_index` is so strange in java, suggest rename `setAppId` and `setPartitionIndex`

src/main/java/com/xiaomi/infra/pegasus/operator/client_operator.java
```diff
@@ -5,33 +5,39 @@
...
     this.pid = gpid;
     this.tableName = tableName;
     this.rpc_error = new error_code();
   }
 
   public client_operator(gpid gpid, String tableName, long partitionHash) {
     this(gpid, tableName);
-    this.header.partition_hash = partitionHash;
+    this.meta.setPartition_hash(partitionHash);
```
Shuo-Jia: same above.

PR #85, opened 2020-01-06
fix: prevent session to be actively closed when it gets response from server

neverchanje:


src/main/java/com/xiaomi/infra/pegasus/rpc/async/SessionFailureDetector.java
```diff
@@ -0,0 +1,40 @@
...
+ * SessionFailureDetector detects whether the session is half-closed by the remote host, in which
+ * case we need to actively close the session and reconnect.
+ */
+public class SessionFailureDetector {
+
+  // Timestamp of the first timed out rpc.
+  private AtomicLong firstRecentTimedOutMs;
+
+  // Session is marked failure if all the RPCs across
+  // `sessionResetTimeWindowMs` are timed out.
```
qinzuoyan, line 15: sessionResetTimeWindowMs -> FAILURE_DETECT_WINDOW_MS

src/main/java/com/xiaomi/infra/pegasus/rpc/async/SessionFailureDetector.java
```diff
@@ -0,0 +1,40 @@
...
+  public static final long FAILURE_DETECT_WINDOW_MS = 10 * 1000; // 10s
```
qinzuoyan, line 16: maybe 10s is too short? consider set it to 30s or 60s?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
```diff
@@ -299,24 +298,18 @@ void tryNotifyFailureWithSeqID(int seqID, error_types errno, boolean isTimeoutTa
       if (!isTimeoutTask && entry.timeoutTask != null) {
```
levy5307, line 298: remove `thows Exception` for `tryNotifyFailureWithSeqID`

src/main/java/com/xiaomi/infra/pegasus/rpc/async/SessionFailureDetector.java
neverchanje: I keep it as it was in this refactoring. I will consider a larger window if it makes sense.

Issue #84, opened 2020-01-06
support backup request on client side


levy5307:
The proposal resides in https://github.com/XiaoMi/pegasus/issues/251.

****一、背景****
--------

Backup request功能可以优化业务在服务抖动时读延迟的长尾问题，适合对一致性要求低的用户。

****二、设计****
------------

1.  增加一个配置项，让用户选择是否启用backup request功能。
2.  在ReplicaConfiguration中添加Map<rpc_address, ReplicaSession> allSessions成员，保存所有的session。
3.  在TableHandler::initTableConfiguration中：
    1.  如果没有开启backup request，则只创建primary副本的session连接放入allSessions。
    2.  如果开启了backup reqeust，secondary也同样要创建连接, 将所有创建的session存入allSessions。
4.  在TableHandler::call中：
    1.  如果是写操作，则要沿用已有逻辑，将请求发往primary。
    2.  如果是读操作，则向allSessions中的所有session都发送请求, 只接收最快的response，其他都忽略（backup reqeust未开启时，allSessions中只有primary session，也相当于只向primary发送）。
        1.  增加is_success(boolean类型), 初始状态下为false，表示请求还没有成功返回，并将其传递给TableHandler::onRpcReply。
        2.  如果isEmpty()，说明还没有和该partition建立好连接，则tryQueryMeta并重新建立连接。
5.  在TableHandler::onRpcReply中，处理请求response：
    1.  如果is_success=true，则说明之前已经有response返回了，则直接忽略。
    2.  如果is\_success=false, 说明该response是第一个返回的, 则按照之前的逻辑处理该请求，处理完之后设置is\_success=true。                                                                               

备注：
*   需要对is_success的处理进行加锁。
*   需要对client_operator中添加一个字段，标记是读操作还是写操作。
*   在4和5中，对于是否开启了backup request是透明的，无需添加额外逻辑判断。只需要在3中创建session时做一下判断。

neverchanje:
> 如果是读操作，则向allSessions中的所有session都发送请求, 只接收最快的response，其他都忽略（backup reqeust未开启时，allSessions中只有primary session，也相当于只向primary发送）。

> 增加is_success(boolean类型), 初始状态下为false，表示请求还没有成功返回，并将其传递给TableHandler::onRpcReply。

> 如果isEmpty()，说明还没有和该partition建立好连接，则tryQueryMeta并重新建立连接。

Could you translate the core ideas in English?

[...]

PR #93, opened 2020-03-05
feat: implement backup request

levy5307:
****Description****

A client first sends one request to the replica believed to be the most appropriate, but then falls back on sending a secondary request after the first request has been outstanding for more than the 95th-percentile(or 99th-percent line, etc) expected latency. The client cancels the remaining outstanding requests once the first result is received. This approach limits the additional load to approximately 5%(1%) while substantially shortening the latency tail.

****How To Use****

We add a new interface in PegasusClient, which name is openTable. Here is the definition:
```
  public PegasusTableInterface openTable(String tableName, int backupRequestDelayMs) throws PException;
```
The second param, which name is backup_request_delay_ms, is the value you want the backup requests delay to sent. If `backup_request_delay_ms <= 0`, the backup request is disabled. 
Please don`t set it to a very small value, because it will increase server load. We strongly suggest to set it to p999(or p99, similarly).  

Note that the old interface of openTable will disable backup request by setting `backupRequestDelayMs = 0` in default.
```
  public PegasusTableInterface openTable(String tableName) throws PException;
```

****Performance Test****

config params
```
field_length(the length value in bytes) = 5000                                 
record_count(the count of write operations) = 300000000                   
thread_count(count of thread) = 50                                 
operation_count(the count of read operations) = 100000000               
readproportion(the proportion of read) = 3                                  
insertproportion(the proportion of write) = 1                                
```

without backup requst:
```
[READ], Operations, 7.4999424E7
[READ], AverageLatency(us), 1055.9526848766197
[READ], MinLatency(us), 75.0
[READ], MaxLatency(us), 2402303.0
[READ], 95.0thPercentileLatency(us), 303.0
[READ], 99.0thPercentileLatency(us), 8359.0
[READ], 99.9thPercentileLatency(us), 208639.0
[READ], 99.99thPercentileLatency(us), 657919.0
[READ], Return=OK, 74999424
```
with backup request (send backup requests after delay 8000 us)
```
[READ], Operations, 7.500155E7
[READ], AverageLatency(us), 1389.2260695278965
[READ], MinLatency(us), 70.0
[READ], MaxLatency(us), 1725439.0
[READ], 95.0thPercentileLatency(us), 8223.0
[READ], 99.0thPercentileLatency(us), 8671.0
[READ], 99.9thPercentileLatency(us), 32751.0
[READ], 99.99thPercentileLatency(us), 136063.0
[READ], Return=OK, 75001550
```

The performance test shows that the p9999 and p999 are go down 5 times than no backup request, with the delay of backup request setting to 8000us(which is p99). But the average latency and p95 is increaced, so backup request will increase server load

https://github.com/XiaoMi/pegasus/issues/251
https://github.com/XiaoMi/pegasus-java-client/issues/84


src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
```diff
@@ -213,6 +221,18 @@ public Builder falconPushInterval(Duration falconPushInterval) {
...
     }
 
+    /**
+     * The delay time of enable backup request. Defaults to {@literal -1}, which means there is no
+     * need to send backup request, see {@link #DEFAULT_BACKUP_REQUEST_DELAY_MS}.
+     *
+     * @param backupRequestDelayMS backupRequestDelayMS
+     * @return {@code this}
+     */
+    public Builder backupRequestDelayMS(int backupRequestDelayMS) {
```
neverchanje: `backupRequestDelayMS` should be a table configuration rather than a client configuration. Because some tables may have strict latency requirements, but some may not.

configuration/pegasus.properties
```diff
@@ -4,3 +4,4 @@ async_workers = 4
 enable_perf_counter = false
 perf_counter_tags = cluster=onebox,app=unit_test
 push_counter_interval_secs = 10
+backup_request_delay_ms = -1
```
Shuo-Jia: “-1” is mean？

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
```diff
@@ -130,6 +133,14 @@ public static Class getSocketChannelClass() {
...
   }
 
+  public int getBackupRequestDelayMS() {
+    return backupRequestDelayMS;
+  }
+
+  public boolean isEnableBackupRequest() {
+    return backupRequestDelayMS > 0;
+  }
+
```
Shuo-Jia: So actually, "0" is also disable? This is different from previous definitions.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -118,22 +119,36 @@ void initTableConfiguration(query_cfg_response resp) {
...
+                if (!secondary.isInvalid()) {
+                  ReplicaSession session = manager_.getReplicaSession(secondary);
+                  s.secondarySessions.add(session);
+                  ChannelFuture channelFuture = session.tryConnect();
+                  if (channelFuture != null) {
+                    futureGroup.add(channelFuture);
+                  }
+                }
+              });
         }
```
Shuo-Jia: `primarySessons` and `secondarySessions` actually is same node, so you shouldn't reconnect.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -211,8 +226,25 @@ void onRpcReply(
       int tryId,
       ReplicaConfiguration cachedHandle,
       long cachedConfigVersion) {
-    client_operator operator = round.getOperator();
+    // judge if it is the first response
+    if (round.isSuccess) {
+      return;
```
Shuo-Jia: `return` will result in the all follow  logics is not executed.(for example, the log will not be printed)

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -238,7 +270,7 @@ void onRpcReply(
         logger.warn(
             "{}: replica server({}) doesn't serve gpid({}), operator({}), try({}), error_code({}), need query meta",
             tableName_,
-            cachedHandle.session.name(),
+            cachedHandle.primarySession.name(),
             operator.get_gpid().toString(),
```
Shuo-Jia: Why only `primarySession` log?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -277,7 +309,11 @@ void onRpcReply(
       tryQueryMeta(cachedConfigVersion);
     }
 
-    tryDelayCall(round, tryId + 1);
+    // must use new round here, because round.isSuccess is true now
+    tryDelayCall(
```
Shuo-Jia: `round.setSuccess(flase)` and then `tryDelayCall(round, tryId + 1)` is not OK?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
levy5307: What is the previous definitions?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
levy5307: yes, we only need the response which is return first. others will be ignored, or the logics will be a mess

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
levy5307: What others do you want to add here? I think infos about primary session is enough here. For most of user requests, secondaries are transparent to them, because they make backup request disabled.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
levy5307: No, we can`t use ```round.setSuccess(flase)```, this will introduce some unexpectable results. For Example:
1. we send backup request to all primary and secondaries.
2.the first response is come back, but there is some wrong with this call, so query meta and set success = false, using ```tryQueryMeta(cachedConfigVersion);``` and  ```round.setSuccess(flase)```
3.the second response is come back,there is also some wrong with it, so it  query meta a second time

So here, I want the response which is first come back, or the logics is in a mess, which will introduce bad maintenance

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
Shuo-Jia: >  /**
     * The delay time of enable backup request. Defaults to {@literal -1}, which means there is no
     * need to send backup request, see {@link #DEFAULT_BACKUP_REQUEST_DELAY_MS}.
     *
You mean -1 is no need, and here means <= 0 is no need?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
levy5307: What did you mean about "same" node? primary and secondary must be not in a same server, in consideration of high avialable.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
Shuo-Jia: I mean if success, you still need excute the follow logic. for example, the logic:
```
switch (operator.rpc_error.errno) {
      case ERR_OK:
        round.thisRoundCompletion();
        return;
.....
```
shouldn't be ignored.

Shuo-Jia: > the response which is return

`round.thisRoundCompletion()` can result callback and return the response. "only return" will not get any response.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
Shuo-Jia: if primary requst failed, we will get the "erro" log, but backup-requst is success. I think we need the primary and secondary request result log

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
levy5307: I have already changed the code, and in my definition, <=0 is disable.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
Shuo-Jia: > same node

Oh, I didn't make it clear. actually is  `the same node-ip connection`. you only need connect all node, the session can be used for primary and secondary.

Shuo-Jia: Well, I see

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
Shuo-Jia: Okey~

src/main/java/com/xiaomi/infra/pegasus/operator/client_operator.java
```diff
@@ -10,18 +10,19 @@
 import com.xiaomi.infra.pegasus.thrift.TException;
 
 public abstract class client_operator {
-  public client_operator(gpid gpid, String tableName) {
+  public client_operator(gpid gpid, String tableName, boolean isWrite) {
```
neverchanje:
```java
class ClientOperatorTypeHelper {
  static bool isWrite(final rrdb_check_and_set_operator op) {
    return true;
  }
  static bool isWrite(final rrdb_get_operator op) {
    return false;
  }
  static bool isWrite(final rrdb_put_operator op) {
    return true;
  }
}
```
I think this better.

levy5307: Yes this way need few modify, but if we want to add a new operator, we also should add a function in ClientOperatorTypeHelper. If somebody forget to add this function, that will introduce a disater.  conversely, if we use write flag in operator, one must set it, or it will produce compile error.
Moreover, I think write or read is a feature of operator, using another class to manage it is a little tricky. Besides, add a flag is once and for all.

acelyc111:
How do I know backup request is enabled or not, could you please add a log when open table?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -118,22 +122,36 @@ void initTableConfiguration(query_cfg_response resp) {
...
+        // backup request is enabled, get all secondary sessions
+        s.secondarySessions.clear();
+        if (isEnableBackupRequest()) {
+          // secondary sessions
+          pc.secondaries.forEach(
+              secondary -> {
+                if (!secondary.isInvalid()) {
+                  ReplicaSession session = manager_.getReplicaSession(secondary);
+                  s.secondarySessions.add(session);
+                  ChannelFuture channelFuture = session.tryConnect();
```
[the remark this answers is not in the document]
levy5307: In the same scope, we cannot define same same for two variables.
And we cann`t use one variable for them, because variable in java is reference to a value, just like `ChannelFuture &fut` in c++

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -118,22 +122,36 @@ void initTableConfiguration(query_cfg_response resp) {
...
+      // create primary sessions
+      s.primaryAddress = pc.primary;
+      if (!pc.primary.isInvalid()) {
+        s.primarySession = manager_.getReplicaSession(pc.primary);
+        ChannelFuture fut = s.primarySession.tryConnect();
+        if (fut != null) {
+          futureGroup.add(fut);
+        }
+
+        // backup request is enabled, get all secondary sessions
```
neverchanje: Don't repeat; Use function, for example:

```java
public ReplicaSession tryConnect(final rpc_address addr, final oldSession, FutureGroup<Void> futureGroup) {
  if(addr.isInvalid()) {
    return null;
  }
  if(oldSession != null && oldSession.getAddress().equals(addr) {
    return oldSession;
  }
  ReplicaSession session = manager_.getReplicaSession(addr);
  ChannelFuture fut = s.primarySession.tryConnect();
  if (fut != null) {
    futureGroup.add(fut);
  }
}
```

```java
        s.primarySession = tryConnect(pc.primary, s.primarySession, futureGroup);
        if (!isEnableBackupRequest()) {
          continue;
        }
        pc.secondaries.forEach(
          secondary -> {
             ReplicaSession session = tryConnect(secondary, null, futureGroup);
             s.secondarySessions.add(session);
          }
      }
```

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -307,16 +347,46 @@ void call(final ClientRequestRound round, final int tryId) {
...
+      if (!round.operator.isWrite && isEnableBackupRequest()) {
+        round.backupRequstTask =
+            executor_.schedule(
+                new Runnable() {
+                  @Override
+                  public void run() {
+                    for (ReplicaSession session : handle.secondarySessions) {
+                      session.asyncSend(
+                          round.getOperator(),
+                          new Runnable() {
```
neverchanje: Prefer to use lambda in this place in order to reduce embed levels.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -307,16 +346,44 @@ void call(final ClientRequestRound round, final int tryId) {
...
+    // send request to primary session
+    if (handle.primarySession != null) {
+      // if it's not write operation and backup request is enabled, schedule to send to secondaries
+      if (!round.operator.isWrite && isBackupRequestEnabled()) {
+        round.backupRequstTask =
+            executor_.schedule(
+                new Runnable() {
+                  @Override
+                  public void run() {
+                    for (ReplicaSession session : handle.secondarySessions) {
```
neverchanje: Your implementation is to emit two backup RPCs to both secondaries. This is not consistent with what you designed.
Here we should choose randomly one secondary only:

>  but then falls back on sending a secondary request after the first request has been outstanding for more than the 95th-percentile(or 99th-percent line, etc) expected latency.

Shuo-Jia: I suggest take turns choosing one secondary but randomly.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
levy5307: At very beginning, I used lambda. But after format, these code is like this:
```
              handle.secondarySessions.forEach(
                  session ->
                      session.asyncSend(
                          round.getOperator(),
                          new Runnable() {
                            @Override
                            public void run() {
                              onRpcReply(round, tryId, tableConfig.updateVersion, session.name());
                            }
                          },
                          round.timeoutMs,
                          true));
```
I think use for loop is more user-friendly:
```
                    for (ReplicaSession session : handle.secondarySessions) {
                      session.asyncSend(
                          round.getOperator(),
                          new Runnable() {
                            @Override
                            public void run() {
                              onRpcReply(round, tryId, tableConfig.updateVersion, session.name());
                            }
                          },
                          round.timeoutMs,
                          true);
                    }
```

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
Shuo-Jia: I suggest take turns but not randomly.

levy5307: I've been waiting for this discussion for a long time. Because if we send to all of these secondaries, we can get the fastest reply from all of these replicas, but the server load will increase(but not too much, because we send to secondaries only when the primary has been outstanding for more than the 95th percentline). So I think this is a trade off.

Shuo-Jia: I remember paper is `one secondary`, or we can add switch to choose `one` or `two`

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
Shuo-Jia: `ReplicaSession session : handle.secondarySessions` ->`ReplicaSession secondarySession : handle.secondarySessions`

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClientRequestRound.java
```diff
@@ -19,6 +20,9 @@
   boolean enableCounter;
   long createNanoTime;
   long expireNanoTime;
+  boolean isCompleted;
+  ScheduledFuture<?> backupRequstTask;
```
neverchanje: backupRequstTask => backupRequestTask

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
levy5307: I will do performance tests to compare these two ways.

levy5307: There is a compromise between the two implementations. We can send a backup request to one of these secondaries(random selection) after delay `backup_request_delay_time`. And wait for the response of them to arrive. If they donot come back after `backup_request_delay_time`, send to another secondary. In the other words, we emit request to the last secondary after delay `2 * backup_request_delay_time`.

neverchanje: @Shuo-Jia 

Doubling the secondaries for backup-request also doubling the risk of cascading failure, when Pegasus has been overwhelmed with a high load (10K QPS for example), the additional load brought by backup-request will be ~20K QPS. Because all the requests fail to complete in expected P99 latency under this load.

Furthermore, we should consider how to prevent cascading failure brought by backup request. My idea is to statistic the ratio of "backup request" and "normal request", and

src/main/java/com/xiaomi/infra/pegasus/operator/rrdb_get_scanner_operator.java
```diff
@@ -17,7 +17,7 @@ public rrdb_get_scanner_operator(
       String tableName,
       get_scanner_request request,
       long partitionHash) {
-    super(gpid, tableName, partitionHash);
+    super(gpid, tableName, partitionHash, false);
```
neverchanje: I suggest uses `allowBackupRequest` instead of `isWrite`. Because Scan is not appropriate for another secondary request. `rrdb_clear_scanner_operator` too.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
Shuo-Jia: > statistic the ratio of "backup request" and "normal request", and if
> the ratio is higher than 5%

`statistic` may be a slightly complicated, for example, you need add qps counter such as server-side and calc the ratio every time?  if you want the  ratio of `backup request` is below 5%, can we send `backup request `by proportional sampling?

neverchanje: Sure. It's merely an idea.

src/main/java/com/xiaomi/infra/pegasus/operator/rrdb_get_scanner_operator.java
levy5307: OK. What about multi-get and batch-get? Do they need backup request?

neverchanje: - multiget 
- get (get includes batch-get)
- ttl
- exist

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1761,7 +1761,9 @@ public void handleReplicaException(
     String replicaServer;
     try {
       replicaServer =
-          replicaConfiguration.primary.get_ip() + ":" + replicaConfiguration.primary.get_port();
+          replicaConfiguration.primaryAddress.get_ip()
+              + ":"
+              + replicaConfiguration.primaryAddress.get_port();
     } catch (UnknownHostException e) {
```
Shuo-Jia: diretly use `toString()`

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1761,7 +1761,9 @@ public void handleReplicaException(
     String replicaServer;
     try {
       replicaServer =
-          replicaConfiguration.primary.get_ip() + ":" + replicaConfiguration.primary.get_port();
+          replicaConfiguration.primaryAddress.get_ip()
+              + ":"
+              + replicaConfiguration.primaryAddress.get_port();
```
Shuo-Jia: diretly use `toString()`

levy5307: I don't want to change these codes in this pr. Maybe we can change it in later pr.

PR #95, opened 2020-03-17
feat: forbid large-size-value written to pegasus server

Shuo-Jia:
Related PR: https://github.com/XiaoMi/rdsn/pull/414. This PR expect to forbid it at client side.
#### New config in `pegasus.properties`:
`enable_write_limit = <true | false>`


src/main/java/com/xiaomi/infra/pegasus/rpc/Cluster.java
```diff
@@ -55,13 +58,19 @@ public static Cluster createCluster(Properties config) throws IllegalArgumentExc
         Integer.parseInt(
             config.getProperty(
                 PEGASUS_PUSH_COUNTER_INTERVAL_SECS_KEY, PEGASUS_PUSH_COUNTER_INTERVAL_SECS_DEF));
+
+    int maxAllowedWriteSize =
+        Integer.parseInt(
+            config.getProperty(PEGASUS_MAX_ALLOWED_WRITE_SIZE, PEGASUS_MAX_ALLOWED_WRITE_SIZE_DEF));
```
levy5307: Should we provide a config to let user set the max allowed write size?

Shuo-Jia: if no config, it will use default value: 1MB

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -389,6 +389,14 @@ public void asyncOperate(client_operator op, ClientOPCallback callback, int time
 
     ClientRequestRound round =
         new ClientRequestRound(op, callback, manager_.counterEnabled(), (long) timeoutMs);
+
+    if (op.enableSizeLimit
+        && manager_.getMaxAllowedWriteSize() > 0
+        && round.operator.header.body_length > manager_.getMaxAllowedWriteSize()) {
+      round.operator.rpc_error.errno = error_types.ERR_INVALID_DATA;
```
levy5307: What's `ERR_INVALID_DATA` mean? Do it same with the error code that rdsn will return when the body size is over allowed write size?

Shuo-Jia: Yeah, it same to rdsn.

src/main/java/com/xiaomi/infra/pegasus/operator/client_operator.java
```diff
@@ -10,26 +10,22 @@
...
+      gpid gpid, String tableName, long partitionHash, boolean enableBackupRequest) {
     this.header = new ThriftHeader();
     this.meta = new request_meta();
     this.meta.setApp_id(gpid.get_app_id());
     this.meta.setPartition_index(gpid.get_pidx());
+    this.meta.setPartition_hash(partitionHash);
     this.pid = gpid;
     this.tableName = tableName;
     this.rpc_error = new error_code();
     this.enableBackupRequest = enableBackupRequest;
```
levy5307: add 'this.enableSizeLimit = false' to imply the enableSizeLimit is fase by default

Shuo-Jia: Java `boolean` default value is `false`, so I don't add it explicitly

neverchanje:
I think it's enough to provide a simple option "enableWriteSizeLimit". If there's some sort of legacy/compatibility problem, the user can disable it and fix the problem in the future.

DynamoDB enforces a maximum size 400KB per item, and 16MB per batch. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html. In terms of Pegasus, we can set limits on set and multiset separately.

src/test/java/com/xiaomi/infra/pegasus/rpc/async/MetaSessionTest.java
```diff
@@ -306,7 +307,7 @@ public void testDNSResetMetaMaxQueryCount() {
   public void testDNSMetaUnavailable() {
     // Ensures when the DNS returns meta all unavailable, finally the query will timeout.
     ClusterManager manager =
-        new ClusterManager(1000, 4, false, null, 60, new String[] {"localhost:34601"});
+        new ClusterManager(1000, 4, false, null, 60, new String[] {"localhost:34601"}, false);
```
neverchanje, line 310: Too many params in the constructor of `ClusterManager`, we can refactor this class later using [Builder Pattern](https://en.wikipedia.org/wiki/Builder_pattern).

src/main/java/com/xiaomi/infra/pegasus/rpc/Table.java
```diff
@@ -52,6 +52,8 @@ public final int getAppID() {
 
   public abstract int getPartitionCount();
 
+  public abstract boolean isEnableWriteLimit();
```
neverchanje, line 55: isWriteLimitEnabled

src/main/java/com/xiaomi/infra/pegasus/tools/WriteLimiter.java
```diff
@@ -0,0 +1,170 @@
...
+    if (sortKey == null) {
+      sortKey = "".getBytes();
+    }
+
+    if (value == null) {
+      value = "".getBytes();
+    }
+
+    if (value.length > SINGLE_VALUE_SIZE) {
+      throw new Exception(
```
neverchanje, line 125: IllegalArgumentException

src/main/java/com/xiaomi/infra/pegasus/rpc/Cluster.java
```diff
@@ -55,17 +58,25 @@ public static Cluster createCluster(Properties config) throws IllegalArgumentExc
         Integer.parseInt(
             config.getProperty(
                 PEGASUS_PUSH_COUNTER_INTERVAL_SECS_KEY, PEGASUS_PUSH_COUNTER_INTERVAL_SECS_DEF));
+
+    boolean enableWriteLimit =
```
neverchanje, line 62: write-limit is not necessary to be initialized for each cluster, it can be initialized as a singleton. What about this way:

```java
class WriteLimitUtils 
{
  public static void validateSingleSet(...) throws IllegalArgumentException {
    if (isWriteLimitEnabled) {
      return;
    }
    ...
  }
  public static void validateMultiSet(...) throws IllegalArgumentException;
  public static void enableWriteLimit() {
    enableWriteLimit = true;
  }
  private static boolean enableWriteLimit = false;
}
```

Shuo-Jia: Well, it's better!

neverchanje, line 62: You actually don't have to modify Cluster/ClusterManager and Table/TableHandler. Try out this way:

```java
class PegasusClient {
  public PegasusClient(Properties config) throws PException {
    this.config = config;
    this.cluster = Cluster.createCluster(config);
    this.enableWriteLimit = Boolean.parseBoolean(
            config.getProperty(PEGASUS_ENABLE_WRITE_LIMIT, PEGASUS_ENABLE_WRITE_LIMIT_DEF));
  }
  public bool isWriteLimitEnabled() { return enableWriteLimit; } 
}
```

```java
class PegasusTable {
  public PegasusTable(PegasusClient client, Table table) {
    this.writeLimiter = new WriteLimiter(client.isWriteLimitEnabled());
  }
}
```

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -604,6 +621,13 @@ public void onCompletion(client_operator clientOP) {
       return promise;
     }
 
+    try {
```
neverchanje, line 624:
```
promise.setFailure(new PException("Exceed write limit threshold:" + e.getMessage()));
```
This line repeats for every operation. Wrap it into a util:

```
  private void limitWrite(Promise promise, IllegalArgumentException exp) {
    if(exp != null) {
      promise.setFailure(new PException("Exceed write limit threshold:" + exp.getMessage()));
    }
  }
```

src/main/java/com/xiaomi/infra/pegasus/rpc/Cluster.java
Shuo-Jia:
```
public PegasusClient(Properties config) throws PException {
    this.config = config;
    this.cluster = Cluster.createCluster(config);
    this.enableWriteLimit = Boolean.parseBoolean(
            config.getProperty(PEGASUS_ENABLE_WRITE_LIMIT, PEGASUS_ENABLE_WRITE_LIMIT_DEF));
  }
I think, the code placed here let our `config` init style is inconsistence and it looks like a little `disharmony`
```

neverchanje: I don't think it really matters. If an option which is only used in PegasusTable, why need passing it to other classes?

src/test/java/com/xiaomi/infra/pegasus/rpc/async/MetaSessionTest.java
Shuo-Jia: Sometime, Builder Pattern isn't also a good choice which is tedious, especially new config need add. I think the viewpoint of first answer: “[builder](https://www.zhihu.com/question/326142180/answer/697172067)” may be referenced?

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
Shuo-Jia: No need to return `Exception`, only wrap it like `handleReplicaException`:
https://github.com/XiaoMi/pegasus-java-client/blob/5d57146f73a0d2d014cdf23c5d39aedb35f263ba/src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java#L1755-L1756

neverchanje:
```
    try {
      writeLimiter.validateMultiSet(hashKey, values);
    } catch (IllegalArgumentException e) {
      handleWriteLimiterException(promise, e.getMessage());
      return promise;
    }
```

compares to

```java
limitWrite(writeLimiter.validateMultiSet(hashKey, values));
```

See which one is better?

Shuo-Jia: `promise` need return if `writeLimiter.validateMultiSet(hashKey, values)` not return null, so the code seem should be:
```java
  IllegalArgumentException exp = writeLimiter.validateMultiSet(hashKey, values);
  limitWrite(promise, exp);
  if (exp != null) {
    return promise
  }
```
or
```java
  IllegalArgumentException exp = writeLimiter.validateMultiSet(hashKey, values);
  if(exp != null){
    limitWrite(promise, exp);
    return promise
  }
```
and java has try...cath, why return `exception` in one method and  judge the `Exception` whether `null` to handle the logic.

PR #96, opened 2020-03-23
feat: handle scan iteration exceed limit and refactor code

hycdong:
As issue [pegasus issue #486](https://github.com/XiaoMi/pegasus/issues/486) and [pegasus pull request #500](https://github.com/XiaoMi/pegasus/pull/500) shows, server limit long time rocksdb iteration, this pull request aims to handle it and refactor scan code.

When response.error = 7 means that client got rocksdb incomplete error, old client will throw exception. However, we should return data have already got and then throw exception.

**Test**
It is hard to construct scan timeout in unit test. As a result, I test this function manually in onebox.
I started onebox with different `iteration_threshold_time`, wrote 10000 kv-pairs in the same hashkey, and test `sortKeyCount`, `multiget`, scanner interface, all results are as expected.

 **Compatibility**

old client -> old server      **ok**
old client -> new server     **ok**
new client -> old server     **ok**
new client -> new server    **ok**

Those situations above have already been tested working well. If old client access new server, it will throw exception immediately when got rocksdb incomplete error and this exception is hard to read, it won't lead any error and compatibility problem.


src/main/java/com/xiaomi/infra/pegasus/client/PegasusScanner.java
```diff
@@ -263,24 +281,24 @@ private void asyncNextInternal() {
...
-  private int _hash_p;
+  private gpid[] _splitGpid;
+  private long[] _splitHash;
+  private int _splitCount;
 
   private gpid _gpid;
   private long _hash;
 
   private List<key_value> _kvs;
   private int _p;
```
neverchanje, line 299: rename `_p` if possible. call it `readKvIndex` or something.

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1665,49 +1665,51 @@ public PegasusScannerInterface getScanner(
...
     }
 
     // check if range is empty
     int cmp = PegasusClient.bytesCompare(start, stop);
 
-    long[] hash;
-    gpid[] v;
-    if (cmp < 0 || cmp == 0 && o.startInclusive && o.stopInclusive) {
+    long[] hashArray;
+    gpid[] gpidArray;
```
neverchanje, line 1701:
```suggestion
    long[] partitionHashes;
    gpid[] partitions;
```

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1720,33 +1722,35 @@ public PegasusScannerInterface getScanner(
...
-    for (int i = 0; i < split; i++) {
-      int s = i < more ? size + 1 : size;
-      gpid[] v = new gpid[s];
-      long[] hash = new long[s];
-      for (int j = 0; j < s; j++) {
-        --count;
-        v[j] = all[count];
-        hash[j] = count;
+    gpid[] allGpid = table.getAllGpid();
+    int gpidCount = allGpid.length;
```
neverchanje, line 1726:
```suggestion
    gpid[] allPartitions = table.getAllGpid();
    int partitionCount = allGpid.length;
```

src/main/java/com/xiaomi/infra/pegasus/client/PegasusScanner.java
```diff
@@ -204,10 +205,15 @@ private void onRecvRpcResponse(error_code err, scan_response response) {
       if (response.error == 0) { // ERR_OK
         _kvs = response.kvs;
         _p = -1;
-        _context = response.context_id;
+        _contextId = response.context_id;
       } else if (response.error
           == 1) { // rocksDB error kNotFound, that scan context has been removed
-        _context = CONTEXT_ID_NOT_EXIST;
+        _contextId = CONTEXT_ID_NOT_EXIST;
+      } else if (response.error == 7) { // rocksDB error kIncomplete
```
neverchanje, line 212: What will happen if the server with scan-limit meets with the client of the previous version? Will the client throw an exception when it accepts `rocksdb::Status::kIncomplete` or just continue?

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1720,33 +1722,35 @@ public PegasusScannerInterface getScanner(
...
+    ScanOptions scanOption = new ScanOptions(options);
+    scanOption.startInclusive = true;
+    scanOption.stopInclusive = false;
+
+    /*
+     * For example, if gpidCount = 16, maxSplitCount = 9
+     * then scannerCount = 9, averageSize = 1, more = 7
+     * It means that we should return nine scanners
+     * the first seventh scanners will contain two partitions' data
+     * and the remainder two scanners will contain only one partition's data
```
neverchanje, line 1742:
```suggestion
     * the first seven scanners will contain two partitions' data
     * and the remaining two scanners will contain only one partition's data
```

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1720,33 +1722,35 @@ public PegasusScannerInterface getScanner(
...
-        --count;
-        v[j] = all[count];
-        hash[j] = count;
+    gpid[] allGpid = table.getAllGpid();
+    int gpidCount = allGpid.length;
+    int scannerCount = gpidCount < maxSplitCount ? gpidCount : maxSplitCount;
+    List<PegasusScannerInterface> ret = new ArrayList<PegasusScannerInterface>(scannerCount);
+
+    int averageSize = gpidCount / scannerCount;
+    int more = gpidCount % scannerCount;
```
neverchanje, line 1731:
```suggestion
    int remainder = gpidCount % scannerCount;
```

src/main/java/com/xiaomi/infra/pegasus/client/PegasusScanner.java
hycdong: In current version, client will throw an exception when it accepts `rocksdb::Status::kIncomplete` immediately and can not get the result of the scanner. In this version, client will also throw an exception but will return scan result.

PR #98, opened 2020-04-03
refactor: optimize perf-counter

Shuo-Jia:
The perf-counter has some problem:
* Existed many redundant counter, such as `*.cps-15min`,`latency.p50` etc,  this pr delete these.
* No existed suitable tag: user usually don't set the tag config or we don‘t get it though it is set, this pr will add **first ip of meta list** and table name as default tag which also join to the user tag if it is set. **add `meta=ip` tag is because sometimes different meta_server has same table name, so that if you search `table=name` , we will not distinguish it**. The follow is an tag example:

> user tag string: cluster=xiaomi
> the final tag : cluster=xiaomi,meta=ip,table=name

* `enable_perf_counter` default value is set `true`

### DIFF
--- a/src/main/java/com/xiaomi/infra/pegasus/metrics/MetricsPool.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/metrics/MetricsPool.java
@@ -42,7 +42,6 @@ public final class MetricsPool {
     theMetric.tags = getTableTag(name);
     theMetric.value = meter.getMeanRate();
     oneMetricToJson(theMetric, output);
-    output.append(',');
   }
 
   public void genJsonsFromHistogram(String name, Histogram hist, StringBuilder output)
@@ -60,7 +59,6 @@ public final class MetricsPool {
     theMetric.tags = getTableTag(name);
     theMetric.value = s.get999thPercentile();
     oneMetricToJson(theMetric, output);
-    output.append(',');
   }
 
   public static void oneMetricToJson(FalconMetric metric, StringBuilder output)
@@ -84,10 +82,12 @@ public final class MetricsPool {
     SortedMap<String, Meter> meters = registry.getMeters();
     for (Map.Entry<String, Meter> entry : meters.entrySet()) {
       genJsonsFromMeter(entry.getKey(), entry.getValue(), builder);
+      builder.append(',');
     }
 
     for (Map.Entry<String, Histogram> entry : registry.getHistograms().entrySet()) {
       genJsonsFromHistogram(entry.getKey(), entry.getValue(), builder);
+      builder.append(',');
     }
     builder.deleteCharAt(builder.length() - 1);
     builder.append("]");
--- a/src/main/java/com/xiaomi/infra/pegasus/metrics/MetricsPool.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/metrics/MetricsPool.java
@@ -16,14 +16,14 @@ import org.json.JSONObject;
 /** Created by weijiesun on 18-3-9. */
 public final class MetricsPool {
 
-  private final String defaultTag;
+  private final String defaultTags;
 
   public MetricsPool(String host, String tags, int reportStepSec) {
     theMetric = new FalconMetric();
     theMetric.endpoint = host;
     theMetric.step = reportStepSec;
     theMetric.tags = tags;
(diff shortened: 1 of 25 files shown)


acelyc111:
> this pr will add **first ip of meta list** and table name as default tag which also join to the user tag if it is set.

cluster meta list is mutable, how to deal with the case when meta server changed?

Shuo-Jia:
> > this pr will add **first ip of meta list** and table name as default tag which also join to the user tag if it is set.
> 
> cluster meta list is mutable, how to deal with the case when meta server changed?

Yeah, I have considered it. Actually, only `table tag` is included in different clusters, meta ip is apply. now, since we meta list is not host name, I haven't find a way the same as sever client to tag one cluster.

src/main/java/com/xiaomi/infra/pegasus/metrics/MetricsPool.java
```diff
@@ -98,27 +82,28 @@ public String metricsToJson() throws JSONException {
...
     builder.append("]");
 
     return builder.toString();
   }
 
+  private String getTableTag(String counterName) {
+    String[] result = counterName.split("@");
+    if (result.length >= 2) {
+      String tableName = result[1].split("\\.")[0];
+      return defaultTag.equals("") ? "table=" + tableName : defaultTag + ",table=" + tableName;
```
neverchanje, line 102:
```suggestion
      if(theMetric.tags.empty() || !theMetric.tags.contains("table=")) {
        // add tag "table=<tableName>"
        return theMetric.tags + ",table=" + tableName;
      }
```

Shuo-Jia:
New commit has delete meta ip

src/main/java/com/xiaomi/infra/pegasus/metrics/MetricsPool.java
```diff
@@ -98,27 +82,30 @@ public String metricsToJson() throws JSONException {
...
     return builder.toString();
   }
 
+  private String getTableTag(String counterName) {
+    if (defaultTag.contains("table=")) {
+      return defaultTag;
+    }
+    String[] result = counterName.split("@");
+    if (result.length >= 2) {
+      return defaultTag.equals("") ? "table=" + result[1] : defaultTag + ",table=" + result[1];
```
levy5307, line 104: `defaultTag.equals("") ? "table=" + result[1] : (defaultTag + ",table=" + result[1]);`

levy5307, line 104: `return defaultTag + ",table=" + result[1];`

levy5307, line 104: return defaultTag.equals("") ? "table=" + result[1] : (defaultTag + ",table=" + result[1]);

Shuo-Jia:
I will re-test it manually before merged

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClientRequestRound.java
```diff
@@ -34,12 +34,15 @@ public ClientRequestRound(
...
-        cb,
-        enableCounter,
-        System.nanoTime() + timeoutInMilliseconds * 1000000L,
-        timeoutInMilliseconds);
+    this.operator = op;
+    this.callback = cb;
+    this.timeoutMs = timeoutInMilliseconds;
+
+    this.enableCounter = enableCounter;
+    this.createNanoTime = System.nanoTime();
```
Shuo-Jia, line 42: `createNanoTime` was deleted in https://github.com/XiaoMi/pegasus-java-client/pull/105, but it is useful in this pr.

src/main/java/com/xiaomi/infra/pegasus/metrics/MetricsPool.java
```diff
@@ -15,11 +15,15 @@
...
 public final class MetricsPool {
+
+  private final String defaultTag;
+
   public MetricsPool(String host, String tags, int reportStepSec) {
     theMetric = new FalconMetric();
     theMetric.endpoint = host;
     theMetric.step = reportStepSec;
     theMetric.tags = tags;
+    defaultTag = tags;
```
levy5307, line 26: `defaultTags`

PR #102, opened 2020-05-11
feat: add a new openTable interface

levy5307:
Add a new openTable interface, which provides a parameter means timeout of this interface.

### DIFF
--- a/src/main/java/com/xiaomi/infra/pegasus/client/PegasusClientInterface.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/client/PegasusClientInterface.java
@@ -65,6 +65,32 @@ public interface PegasusClientInterface {
       throws PException;
 
   /**
+   * Open a table, and prepare the sessions and route-table to the replica-servers.
+   *
+   * <p>Please notice that pegasus support two kinds of API: 1. the client-interface way, which is
+   * provided in this class. 2. the table-interface way, which is provided by {@link
+   * PegasusTableInterface}. With the client-interface, you don't need to create
+   * PegasusTableInterface by openTable, so you can access the pegasus cluster conveniently.
+   * However, the client-interface's api also has some restrictions: 1. we don't provide async
+   * methods in client-interface. 2. the timeout in client-interface isn't as accurate as the
+   * table-interface. 3. the client-interface may throw an exception when open table fails. It means
+   * that you may need to handle this exception in every data access operation, which is annoying.
+   * 4. You can't specify a per-operation timeout. So we recommend you to use the table-interface.
+   *
+   * @param tableName the table should be exist on the server, which is created before by the system
+   *     administrator
+   * @param backupRequestDelayMs the delay time to send backup request. If backupRequestDelayMs <=
+   *     0, The backup request is disabled.
+   * @param timeout how long will the operation timeout in milliseconds. if timeout > 0, it is a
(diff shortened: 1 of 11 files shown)


src/main/java/com/xiaomi/infra/pegasus/rpc/async/MetaSession.java
```diff
@@ -18,11 +18,7 @@
 
 public class MetaSession extends HostNameResolver {
   public MetaSession(
-      ClusterManager manager,
-      String[] addrList,
-      int eachQueryTimeoutInMills,
-      int defaultMaxQueryCount,
-      EventLoopGroup g)
+      ClusterManager manager, String[] addrList, int defaultMaxQueryCount, EventLoopGroup g)
```
hycdong, line 21: Why you remove `eachQueryTimeoutInMills` and add parameter called timeout for `asyncQuery`?

src/main/java/com/xiaomi/infra/pegasus/client/PegasusClientInterface.java
```diff
@@ -64,6 +64,32 @@
...
+   * <p>Please notice that pegasus support two kinds of API: 1. the client-interface way, which is
+   * provided in this class. 2. the table-interface way, which is provided by {@link
+   * PegasusTableInterface}. With the client-interface, you don't need to create
+   * PegasusTableInterface by openTable, so you can access the pegasus cluster conveniently.
+   * However, the client-interface's api also has some restrictions: 1. we don't provide async
+   * methods in client-interface. 2. the timeout in client-interface isn't as accurate as the
+   * table-interface. 3. the client-interface may throw an exception when open table fails. It means
+   * that you may need to handle this exception in every data access operation, which is annoying.
+   * 4. You can't specify a per-operation timeout. So we recommend you to use the table-interface.
+   *
```
neverchanje, line 79:
```suggestion
```

src/main/java/com/xiaomi/infra/pegasus/rpc/async/MetaSession.java
levy5307: In this pr, we should add some interface in MetaSession, which name are query and asyncQuery, which should have a timeout parameter.
And `eachQueryTimeoutInMills` is used in the old version of query and asyncQuery, which have no timeout parameter. So If we retain `eachQueryTimeoutInMills`, we should retain these old interfaces.
In my completetion, I delete these old interfaces.  We should pass timeout to default value or specific value, when you call these interfaces. In this way, there is no need to retain these two versions of interfaces.

levy5307:
> I'm not sure if adding such simple an interface with so many modifications is necessary. It's a damn ugly design.

Because we should pass the timeout parameter through to metaSession, so we should change relevant code. Do you have another way which should add more elegance?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/TableHandler.java
```diff
@@ -230,7 +230,8 @@ public void run() {
                 onUpdateConfiguration(query_op);
               }
             },
-            5);
+            5,
+            getDefaultTimeout());
```
neverchanje, line 234: But what if he set the default timeout too small? It means even if he initially opens the table at a long timeout, when the table configuration is changed he still has problem on opening table.

PR #104, opened 2020-05-11
refactor: reduce parameters in construction of internal Table and Cluster

neverchanje:
To add a simple configuration to control pegasus-client's behavior we formerly needed to modify many files, including 'Cluster', 'ClusterManager',"Table","TableHandler", and all the unit tests under package "pegasus/rpc/aysnc", since the constructor of the above classes were changed. This is an typical extensibility issue in software design.

To add config without touching the related files, I add "TableOptions" and "ClusterOptions". Adding config will add a field to the options, "enableWriteLimit" for example, but the constructors keep untouched:

```java
public ClusterManager(ClusterOptions opts)
```


src/main/java/com/xiaomi/infra/pegasus/rpc/ClusterOptions.java
```diff
@@ -0,0 +1,116 @@
...
+
+  public static final String PEGASUS_OPERATION_TIMEOUT_KEY = "operation_timeout";
+  public static final String PEGASUS_OPERATION_TIMEOUT_DEF = "1000";
+
+  public static final String PEGASUS_ASYNC_WORKERS_KEY = "async_workers";
+  public static final String PEGASUS_ASYNC_WORKERS_DEF =
+      String.valueOf(Runtime.getRuntime().availableProcessors());
+
+  public static final String PEGASUS_ENABLE_PERF_COUNTER_KEY = "enable_perf_counter";
+  public static final String PEGASUS_ENABLE_PERF_COUNTER_VALUE = "false";
```
levy5307, line 22: `PEGASUS_ENABLE_PERF_COUNTER_VALUE`  --> `PEGASUS_ENABLE_PERF_COUNTER_DEF`

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
```diff
@@ -43,32 +43,25 @@
...
+    replicaGroup = getEventLoopGroupInstance(opts.asyncWorkers());
     metaGroup = getEventLoopGroupInstance(1);
     tableGroup = getEventLoopGroupInstance(1);
 
-    metaList = address_list;
+    metaList = opts.metaList();
     // the constructor of meta session is depend on the replicaSessions,
     // so the replicaSessions should be initialized earlier
-    metaSession = new MetaSession(this, address_list, timeout, 10, metaGroup);
+    metaSession = new MetaSession(this, opts.metaList(), opts.operationTimeout(), 10, metaGroup);
```
levy5307, line 61: I think you can pass options to the constructor of MetaSession

src/main/java/com/xiaomi/infra/pegasus/rpc/Cluster.java
```diff
@@ -8,65 +8,14 @@
...
-        perfCounterTags,
-        pushIntervalSecs,
-        address);
+    return new ClusterManager(ClusterOptions.create(config));
   }
 
   public abstract String[] getMetaList();
 
-  public abstract Table openTable(String name, KeyHasher function, int backupRequestDelayMs)
+  public abstract Table openTable(String name, TableOptions config)
```
neverchanje, line 18: options

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
```diff
@@ -43,32 +43,25 @@
...
-      boolean enableCounter,
-      String perfCounterTags,
-      int pushIntervalSecs,
-      String[] address_list)
-      throws IllegalArgumentException {
-    setTimeout(timeout);
-    this.enableCounter = enableCounter;
+  public ClusterManager(ClusterOptions opts) throws IllegalArgumentException {
+    setTimeout(opts.operationTimeout());
+    this.enableCounter = opts.enablePerfCounter();
```
levy5307, line 48: I think there is no need to saving too many copies for the member of options. We can simply remain a reference to options here.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
neverchanje: Yes but this is not my concern in this PR. We can do this refactoring later when it needs to add new param to MetaSession.

src/main/java/com/xiaomi/infra/pegasus/client/PegasusClient.java
```diff
@@ -72,11 +73,11 @@ private PegasusTable getTable(String tableName, int backupRequestDelayMs) throws
   // pegasus client configuration keys
   public static final String[] PEGASUS_CLIENT_CONFIG_KEYS =
       new String[] {
-        Cluster.PEGASUS_META_SERVERS_KEY,
-        Cluster.PEGASUS_OPERATION_TIMEOUT_KEY,
-        Cluster.PEGASUS_ASYNC_WORKERS_KEY,
-        Cluster.PEGASUS_ENABLE_PERF_COUNTER_KEY,
-        Cluster.PEGASUS_PERF_COUNTER_TAGS_KEY,
+        ClusterOptions.PEGASUS_META_SERVERS_KEY,
```
levy5307, line 76: I think we can add a interface for ClusterOptions to get client config keys.

src/main/java/com/xiaomi/infra/pegasus/client/PegasusClient.java
```diff
@@ -72,11 +73,11 @@ private PegasusTable getTable(String tableName, int backupRequestDelayMs) throws
...
+        ClusterOptions.PEGASUS_OPERATION_TIMEOUT_KEY,
```
levy5307, line 77: I think we can implement a interferce for ClusterOptions to get client config keys.

neverchanje: I can't see it makes any better. Leave it later.

levy5307: there is no need to change PegasusClient.java if we add a option to ClusterOptions.

neverchanje: OK, sounds reasonable.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
neverchanje: Leave it later when it needs to add param to MetaSession.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
neverchanje: Only `operationTimeout`, `metaList`, and `enableCounter` are copied as private members of `ClusterManager`, not "too many". I don't wanna change much. This is not my concern in this PR.

PR #108, opened 2020-05-20
fix: startSortKey can't  pass null

Shuo-Jia:
`startSortKey` should support pass `null` 

```java
  public void delRange(
      byte[] hashKey, byte[] startSortKey, byte[] stopSortKey, DelRangeOptions options, int timeout)
      throws PException {
    ScanOptions scanOptions = new ScanOptions();
    scanOptions.noValue = true;
    scanOptions.startInclusive = options.startInclusive;
    scanOptions.stopInclusive = options.stopInclusive;
    scanOptions.sortKeyFilterType = options.sortKeyFilterType;
    scanOptions.sortKeyFilterPattern = options.sortKeyFilterPattern;

    options.nextSortKey = new String(startSortKey);
```

```java
public class DelRangeOptions {
  public String nextSortKey = "";
```

Because DelRangeOptions.nextSortKey was a string, if startSortKey was passed null, `new String(null)` will throw NullPointerException.

**Besides**, this pr update junit4 to junit5 for simplying unit code, for example, support test `exception` method, see [junit5](https://howtodoinjava.com/junit5/expected-exception-example/)

### DIFF
--- a/src/main/java/com/xiaomi/infra/pegasus/client/DelRangeOptions.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/client/DelRangeOptions.java
@@ -4,7 +4,7 @@
 package com.xiaomi.infra.pegasus.client;
 
 public class DelRangeOptions {
-  public String nextSortKey = "";
+  public byte[] nextSortKey = null;
   public boolean startInclusive = true; // if the startSortKey is included
   public boolean stopInclusive = false; // if the stopSortKey is included
   public FilterType sortKeyFilterType = FilterType.FT_NO_FILTER; // filter type for sort key
--- a/src/test/java/com/xiaomi/infra/pegasus/client/TestBasic.java
+++ b/src/test/java/com/xiaomi/infra/pegasus/client/TestBasic.java
@@ -12,7 +12,9 @@ import java.util.Map;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.junit.Assert;
-import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assertions.*;
+import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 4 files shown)


src/test/java/com/xiaomi/infra/pegasus/client/TestBasic.java
```diff
@@ -2482,6 +2484,41 @@ public void delRange() throws PException {
...
+    remainingValue.clear();
+    valueStr.clear();
+    DelRangeOptions delRangeOptions2 = new DelRangeOptions();
+
+    // test hashKey can't be null or ""
+    try {
+      client.delRange(
+          tableName, "delRange".getBytes(), "k1".getBytes(), "k2".getBytes(), delRangeOptions2);
+    } catch (Exception e) {
+      Assert.assertTrue(e.getMessage().contains("Invalid parameter: hash key can't be empty"));
```
neverchanje, line 2497: I can't see why `"delRange".getBytes()` is an empty hashkey... I think this test passed because it never catches any exception. The best practice to assert an exception is thrown is using `assertThrow`

```java
  Exception e = assertThrows(PException.class, () -> {
      client.delRange(
          tableName, "delRange".getBytes(), "k1".getBytes(), "k2".getBytes(), delRangeOptions2);
  });
  Assert.assertTrue(e.getMessage().contains("Invalid parameter: hash key can't be empty"));
```

https://howtodoinjava.com/junit5/expected-exception-example/

src/main/java/com/xiaomi/infra/pegasus/client/DelRangeOptions.java
```diff
@@ -4,7 +4,7 @@
 package com.xiaomi.infra.pegasus.client;
 
 public class DelRangeOptions {
-  public String nextSortKey = "";
+  public byte[] nextSortKey = null;
```
levy5307, line 7: why did you set nextSortKey = null in default?

PR #109, opened 2020-05-20
build(deps): add dependency of javax.annotation-api

qinzuoyan:


neverchanje:
why add this dependency?

qinzuoyan:
because we used it in https://github.com/XiaoMi/pegasus-java-client/blob/thrift-0.11.0-inlined/src/main/java/com/xiaomi/infra/pegasus/apps/incr_request.java#L13
```
@javax.annotation.Generated(value = "Autogenerated by Thrift Compiler (0.11.0)", date = "2020-04-13")
```

neverchanje:
Ok. That's because JDK11 remove javax.annotation package. https://www.oracle.com/technetwork/java/javase/11-relnote-issues-5012449.html#JDK-8190378. I've included openjdk11 in travis.

PR #110, opened 2020-06-04
feat: support prometheus

Shuo-Jia:


src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
```diff
@@ -45,7 +46,8 @@
   private final Duration operationTimeout;
   private final int asyncWorkers;
   private final boolean enablePerfCounter;
-  private final String falconPerfCounterTags;
+  private final String perfCounterType;
```
levy5307, line 49: what about define a enum to represent the type of perf counter?

PR #113, opened 2020-06-26
fix: release streams resource while loading config

### LINKED ISSUE #112: Unreleased Resource: Streams
https://github.com/XiaoMi/pegasus-java-client/blob/33a9a05f1929293275a1952dbd4457923f83e1f9/src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java#L64

The program can potentially fail to release a system resource.

totalo:
fixed #112

### DIFF
--- a/src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
@@ -78,19 +78,13 @@ public class PConfigUtil {
         throw new PException("config resource not found: " + configPath);
       }
       config.load(stream);
+      stream.close();
       return config;
     } catch (Throwable e) {
       if (e instanceof PException) {
         throw (PException) e;
       } else {
         throw new PException(e);
-      }
-    } finally {
-      try {
-        if (stream != null) {
-          stream.close();
-        }
-      } catch (Exception ignore) {
       }
     }
   }
--- a/src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
@@ -61,11 +61,12 @@ public class PConfigUtil {
             new BufferedInputStream(
                 new FileInputStream(configPath.substring(PConfigUtil.LOCAL_FILE_PREFIX.length()))));
       } else if (PConfigUtil.isResource(configPath)) {
-        try (InputStream stream = PegasusClient.class.getResourceAsStream(
+        try (InputStream stream =
+            PegasusClient.class.getResourceAsStream(
                 configPath.substring(PConfigUtil.RESOURCE_PREFIX.length()))) {
-            config.load(stream);
-          } catch (NullPointerException e) {
-            throw new PException("config resource not found: " + configPath);
+          config.load(stream);
+        } catch (NullPointerException e) {
+          throw new PException("config resource not found: " + configPath);
         }
       } else {
         throw new PException(
--- a/src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
@@ -8,6 +8,7 @@ import java.io.ByteArrayInputStream;
 import java.io.FileInputStream;
 import java.io.InputStream;
 import java.util.Properties;
+
 import org.I0Itec.zkclient.ZkClient;
 import org.I0Itec.zkclient.serialize.BytesPushThroughSerializer;
 import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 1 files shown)


src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
```diff
@@ -61,13 +61,13 @@ public static Properties loadConfiguration(String configPath) throws PException
...
                 new FileInputStream(configPath.substring(PConfigUtil.LOCAL_FILE_PREFIX.length()))));
       } else if (PConfigUtil.isResource(configPath)) {
-        InputStream stream =
+        try (InputStream stream =
             PegasusClient.class.getResourceAsStream(
-                configPath.substring(PConfigUtil.RESOURCE_PREFIX.length()));
-        if (stream == null) {
+                configPath.substring(PConfigUtil.RESOURCE_PREFIX.length()))) {
+          config.load(stream);
+        } catch (NullPointerException e) {
```
neverchanje, line 68: Thank you for your contribution, but I'm afraid this is not a cool way to fix the issue.
Issue #112  indicates that the stream was not closed after usage. I don't see wrapping around with a try-catch can fix this.

The following solution may be better:

```diff
        InputStream stream =
            PegasusClient.class.getResourceAsStream(
                configPath.substring(PConfigUtil.RESOURCE_PREFIX.length()));
        if (stream == null) {
          throw new PException("config resource not found: " + configPath);
        }
        config.load(stream);
+       stream.close();
```

totalo: Ok, Fine, But I want to know why can't we use try-with-resource?

neverchanje: You don't fix anything using try-with-resource AFAIS. There's no reason to change the code if nothing fixed.
What do you think of the benefit that adding try-catch here can bring?

totalo: Thank you,😁 I understand that using it there is not a good way.

src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
```diff
@@ -68,6 +68,7 @@ public static Properties loadConfiguration(String configPath) throws PException
           throw new PException("config resource not found: " + configPath);
         }
         config.load(stream);
+        stream.close();
```
Shuo-Jia, line 71: if here need manual-close, the above `stream` also need close, maybe you can modify:
```java
      InputStream stream = null;
      if (PConfigUtil.isZkPath(configPath)) {
        stream = new ByteArrayInputStream(PConfigUtil.loadConfigFromZK(configPath));
      } else if (PConfigUtil.isLocalFile(configPath)) {
        stream = new BufferedInputStream(
            new FileInputStream(configPath.substring(PConfigUtil.LOCAL_FILE_PREFIX.length())));
      } else if (PConfigUtil.isResource(configPath)) {
        stream =
            PegasusClient.class.getResourceAsStream(
                configPath.substring(PConfigUtil.RESOURCE_PREFIX.length()));
      } else {
        throw new PException(
            "configPath format error, "
                + "should be local file format as 'file:///path/to/config', "
                + "or zookeeper path format as 'zk://host1:port1,host2:port2,host3:port3/path/to/config', "
                + "or java resource format as 'resource:///path/to/config', "
                + "but actual configPath is "
                + configPath);
      }
      if (stream == null) {
        throw new PException("config resource not found: " + configPath);
      }
      config.load(stream);
      return config;
```

totalo: Yeh, Got its!

src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
```diff
@@ -77,13 +74,25 @@ public static Properties loadConfiguration(String configPath) throws PException
...
       if (e instanceof PException) {
         throw (PException) e;
       } else {
         throw new PException(e);
       }
+    } finally {
+      if (stream != null) {
+        try {
+          stream.close();
+        } catch (Exception ignore) {
```
Shuo-Jia, line 92: Better output `debug` info？

src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
```diff
@@ -77,13 +74,25 @@ public static Properties loadConfiguration(String configPath) throws PException
                 + "but actual configPath is "
                 + configPath);
       }
+      if (stream == null) {
+        throw new PException("config resource not found: " + configPath);
+      }
+      config.load(stream);
```
neverchanje, line 80:
```suggestion
      config.load(stream);
      stream.close();
```

Close here. So you don't have to add another try-catch after `finally`.

src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
```diff
@@ -77,13 +74,25 @@ public static Properties loadConfiguration(String configPath) throws PException
...
         throw new PException(e);
       }
+    } finally {
+      if (stream != null) {
+        try {
+          stream.close();
+        } catch (Exception ignore) {
+          // ignore
+        }
+      }
```
neverchanje, line 95:
```suggestion
```

See my comment above

src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
totalo: If `config.load` throws an exception, resources cannot be released normally.😂

src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
totalo: Abnormal shutdown of resources is not the business logic of the system, I think it can be ignored, what do you think?

src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
neverchanje: Sure. But the stream will not be closed before `config.load` works are done.

```suggestion
      config.load(stream); // if any exception, it will be thrown out, won't steps down to `stream.close()`
      stream.close();
```

totalo: Although the `config.load` works is not completed, I think the resource also  needs to be released.

src/main/java/com/xiaomi/infra/pegasus/client/PConfigUtil.java
neverchanje: @totalo  You can throw it as a `PException`. Don't ignore any exception if we can do better.

```diff
-  // ignore
+  throw new PException(e);
```

totalo: Yeh，Got it! Thank you~

PR #117, opened 2020-07-23
refactor: delete duplicate config and use ClientOptions uniformly

Shuo-Jia:
Now the config of `PegasusClient` is duplicate, which existed in `ClusterOptions`(it loads from config file) and `ClientOptions`(it is created manualy), this PR del the `ClusterOptions` and use `ClientOptions` uniformly.


src/main/java/com/xiaomi/infra/pegasus/client/PegasusClient.java
```diff
@@ -25,14 +23,10 @@
 public class PegasusClient implements PegasusClientInterface {
   private static final Logger LOGGER = LoggerFactory.getLogger(PegasusClient.class);
 
-  public static final String PEGASUS_ENABLE_WRITE_LIMIT = "enable_write_limit";
-  public static final String PEGASUS_ENABLE_WRITE_LIMIT_DEF = "true";
+  private ClientOptions clientOptions;
```
levy5307, line 26: should you make `clientOptions` to be a const member?

pom.xml
```diff
@@ -35,6 +35,11 @@
       <artifactId>metrics-core</artifactId>
       <version>3.1.2</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-configuration2</artifactId>
```
neverchanje, line 40: Can this dep be shaded?

Shuo-Jia: Yeah, but If it hasn't compatibility problems, it maybe no need to be shaded, otherwise, should all dependencys be shaded?

neverchanje: I'm afraid so. The deps for non-testing code generally need to be shaded in case of dep conflict.
But given the apache.common.configuration2 is not much used here, we can leave it unshaded for now.

PR #120, opened 2020-07-24
docs: add example on table scan

totalo:


src/main/java/com/xiaomi/infra/pegasus/example/ExampleSample.java
```diff
@@ -0,0 +1,40 @@
...
+  public static void main(String[] args) throws PException, UnsupportedEncodingException {
+    PegasusClientInterface client = PegasusClientFactory.createClient(ClientOptions.create());
+    PegasusTableInterface table = client.openTable("temp");
+
+    byte[] value = new byte[1000];
+
+    for (int i = 0; i < 1000; i++) {
+      value[i] = 'x';
+    }
+
```
Shuo-Jia, line 22: Long value can be generate by `RandomStringUtils.random`

src/main/java/com/xiaomi/infra/pegasus/example/FullScanSample.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * @param tableName
+   * @throws PException
+   * @throws IOException
+   */
+  private static void searchHistoryOneYearAgo(String tableName) throws PException, IOException {
+    PegasusClientInterface client =
+        PegasusClientFactory.createClient(ClientOptions.builder().build());
```
Shuo-Jia, line 42: ClientOptions.create()

src/main/java/com/xiaomi/infra/pegasus/example/ExampleSample.java
totalo: yeah! Got it

Shuo-Jia:
Thanks for your commit, the sample resolve a common case, it's very good. But as SDK API example, I think it should  be as simple as possible, so that user can get the way of api soon, and no need take time to understand the specially case.

So how about simplify the sample code? A suggection is only focus the api usage, delete the specially case？

totalo:
> Thanks for your commit, the sample resolve a common case, it's very good. But as SDK API example, I think it should be as simple as possible, so that user can get the usage of api soon, and no need take time to understand the specially case.
> 
> So how about simplify the sample code? A suggection is only focus the api usage, delete the specially case？And then you may add more necessary api sample in one class.

Ok ， I will try it!

src/main/java/com/xiaomi/infra/pegasus/example/FullScanSample.java
```diff
@@ -0,0 +1,48 @@
...
+      long start = System.currentTimeMillis();
+      while (true) {
+        Pair<Pair<byte[], byte[]>, byte[]> pair = scan.next();
+        if (null == pair) break;
+        Pair<byte[], byte[]> keys = pair.getLeft();
+        byte[] hashKey = keys.getLeft();
+        byte[] sortKey = keys.getRight();
+        cnt++;
+        // get value
+        byte[] value = table.get(hashKey, sortKey, 3000);
```
Shuo-Jia, line 39: `value = pair.getRight()`, no need use table.get. So `PegasusTableInterface table = client.openTable("temp");` should remve

src/main/java/com/xiaomi/infra/pegasus/example/FullScanSample.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public static void main(String[] args) throws PException, UnsupportedEncodingException {
+    PegasusClientInterface client = PegasusClientFactory.createClient(ClientOptions.create());
+    PegasusTableInterface table = client.openTable("temp");
+
+    // Set up the scanners.
+    ScanOptions scanOptions = new ScanOptions();
+    scanOptions.batchSize = 20;
+    // Values can be optimized out during scanning to reduce the workload.
+    scanOptions.noValue = true;
```
Shuo-Jia, line 22: In former case, set `scanOptions.noValue = true` is because that case only need `hashKey/sortKey` to filter result. In this general case, this options may not be appropriate. And I see the follow code you add `table.get` to obtain the `value` which is redundant. I suggest you just explain the `scanOptions.noValue = true`, And use `pair.getRight` to get value in follow code.

totalo: Ok， Got it！ I use `table.get `because `pari.getRight `has no value, I understand now, thanks~

src/main/java/com/xiaomi/infra/pegasus/example/ExampleSample.java
```diff
@@ -0,0 +1,37 @@
...
+package com.xiaomi.infra.pegasus.example;
+
+import com.xiaomi.infra.pegasus.client.*;
+import java.io.UnsupportedEncodingException;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.lang3.RandomStringUtils;
+
+/** Example sample */
+public class ExampleSample {
```
neverchanje, line 13: I think this example is too naive to show any best practice of the SDK. The Full-Scan example is good. But this example I'm afraid has no much value.

A better example must show the advanced cases that this SDK can handle. The hbase-client example:

https://github.com/apache/hbase/tree/master/hbase-examples/src/main/java/org/apache/hadoop/hbase/client/example

I suggest you can remove this sample from this PR. And we can go on merge the full-scan sample.

src/main/java/com/xiaomi/infra/pegasus/example/FullScanSample.java
```diff
@@ -0,0 +1,47 @@
...
+  private static final String tableName = "temp";
+
+  public static void main(String[] args) throws PException, UnsupportedEncodingException {
+    PegasusClientInterface client = PegasusClientFactory.createClient(ClientOptions.create());
+
+    // Set up the scanners.
+    ScanOptions scanOptions = new ScanOptions();
+    scanOptions.batchSize = 20;
+
+    List<PegasusScannerInterface> scans = client.getUnorderedScanners(tableName, 16, scanOptions);
```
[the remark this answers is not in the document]
totalo: Well, I will pay attention to these later!

src/main/java/com/xiaomi/infra/pegasus/example/FullScanSample.java
```diff
@@ -0,0 +1,47 @@
...
+      int cnt = 0;
+      long start = System.currentTimeMillis();
+      while (true) {
+        Pair<Pair<byte[], byte[]>, byte[]> pair = scan.next();
+        if (null == pair) break;
+        Pair<byte[], byte[]> keys = pair.getLeft();
+        byte[] hashKey = keys.getLeft();
+        byte[] sortKey = keys.getRight();
+        cnt++;
+        System.out.printf(
```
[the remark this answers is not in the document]
totalo: Got it！

src/main/java/com/xiaomi/infra/pegasus/example/ExampleSample.java
totalo: Ok, I will do these things later, but there is a problem. I think that these examples should not be included when packaged into a jar package. They should not be included with the core code. What do you think of putting it outside?

neverchanje: Yes. I'm aware of this too. I think it's fine to let it be packaged for now, like HBase. 
Users are not able to directly use the example.

PR #133, opened 2020-09-14
feat(security): implement start negotiation

levy5307:
1. add security.thrift and the corresponding generated classes, for example: negotiation_reqeust, negotiation_response, and so on.
2. add negotiation_operator
3. implement start negotiation
4. add unit test


src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
```diff
@@ -210,14 +223,13 @@ public void operationComplete(ChannelFuture channelFuture) throws Exception {
...
-      negotiation = new Negotiation(this);
-      negotiation.start();
-    } else {
-      logger.info("{}: mark session state connected");
-      markSessionConnected(activeChannel);
-    }
+    VolatileFields newCache = new VolatileFields();
+    newCache.state = ConnState.CONNECTING;
+    newCache.nettyChannel = activeChannel;
+    fields = newCache;
```
Shuo-Jia, line 229: I don’t understand why update `fields`？

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
```diff
@@ -119,6 +121,17 @@ public int asyncSend(
...
 
+  public void sendNegoMsg(negotiation_request msg, long timeoutInMilliseconds) {
+    final RequestEntry entry = new ReplicaSession.RequestEntry();
+    entry.sequenceId = seqId.getAndIncrement();
+    entry.op = new negotiation_operator(msg);
+    entry.callback = new Negotiation.RecvHandler((negotiation_operator) entry.op, this);
+    entry.timeoutTask = addTimer(entry.sequenceId, timeoutInMilliseconds);
+    pendingResponse.put(entry.sequenceId, entry);
+
+    write(entry, fields);
```
Shuo-Jia, line 132: Why re-use `asyncSend`?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
levy5307: if we don't update `fields`, the nettyChannel is null. We need nettyChannel to send negotiation message.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
levy5307: asyncSend need the connect state to be `CONNECTED`, but in the duration of negotiation, the connect state is not `CONNECTED`

Shuo-Jia: And `sendNegoMsg` should be `synchronized` and double-check connect-state to avoid re-send?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/Negotiation.java
```diff
@@ -1,13 +1,79 @@
...
+      if (resp == null) {
+        logger.error("RecvHandler received a null response, abandon it");
+        return;
+      }
+
+      switch (resp.status) {
+        case SASL_LIST_MECHANISMS_RESP:
+        case SASL_SELECT_MECHANISMS_RESP:
+        case SASL_CHALLENGE:
+        case SASL_SUCC:
```
Shuo-Jia, line 62: is `SASL_SUCC` should execute `markSessionConnected`

levy5307: yes

src/main/java/com/xiaomi/infra/pegasus/rpc/async/Negotiation.java
```diff
@@ -1,13 +1,79 @@
...
+
+  public negotiation_status get_status() {
+    return status;
   }
 
   private ReplicaSession session;
+  private negotiation_status status;
+
+  private static final Logger logger = org.slf4j.LoggerFactory.getLogger(Negotiation.class);
+  private static final int defaultTimeoutMs = 3000;
```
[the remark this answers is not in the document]
levy5307: I think 3s is long enough, what's you opinion?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
neverchanje: I think you should not modify any ReplicaSession code to implement negotiation. Implement upon it.
Try to think of any other design alternative. Adding features by intrusive modifications to an existing class is the last thing we should do.

ReplicaSession contains pure RPC logic. The negotiation is essentially based on RPC too. So why break this design? Why add responsibility to ReplicaSession?

One approach is to inherit ReplicaSession to SecureSession and implement negotiation in the subclass.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/Negotiation.java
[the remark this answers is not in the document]
Shuo-Jia: `Negotiation` usually the first rpc call, it may need longer time to ensure success, the metaServer call timeout is 10s? I think it can be same.

PR #125, opened 2020-08-25
refactor: using "interceptor" to implement backup-request

Shuo-Jia:
This PR is implement of `Interceptor` of issue #123 and refeator the backup-request via using it

### manual test
- deploy pegasus 2.0 and packge the commit
- start client `set`、`get` and check client monitor(operation qps) and server monitor(backup qps)
- java client running well and the client monitor and server monitor can be see counter

PR #135, opened 2020-09-21
feat(security): add ReplicaSession interceptor 

levy5307:
* add ReplicaSession hook to deal with some situations, such as session connected、sending a message.
* add some configurations


src/main/java/com/xiaomi/infra/pegasus/rpc/async/SecurityReplicaSessionHook.java
```diff
@@ -0,0 +1,45 @@
...
+    try {
+      loginContext.login();
+    } catch (LoginException le) {
+      logger.error("authentication failed", le);
+      System.exit(-1);
+    }
+
+    subject = loginContext.getSubject();
+    logger.info("login succeed, as user {}", subject.getPrincipals().toString());
+  }
```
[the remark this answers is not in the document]
levy5307: The javax.security.auth.login.LoginContext class provides the basic methods used to authenticate subjects, and provides a way to develop an application independent of the underlying authentication technology. The LoginContext consults a Configuration to determine the authentication services, or LoginModule(s), configured for a particular application. Therefore, different LoginModules can be plugged in under an application without requiring any modifications to the application itself.
(Ref)[https://docs.oracle.com/javase/7/docs/technotes/guides/security/jaas/JAASRefGuide.html#LoginContext]

configuration/pegasus_jaas.conf
```diff
@@ -0,0 +1,10 @@
+client {
+	com.sun.security.auth.module.Krb5LoginModule required
```
neverchanje, line 2: Does this config format support comment? Can you write a comment of description on top of each config item?
Furthermore, does it mean that every of our users should have this file in their project?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSessionHookManager.java
```diff
@@ -0,0 +1,23 @@
+package com.xiaomi.infra.pegasus.rpc.async;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ReplicaSessionHookManager {
```
neverchanje, line 6: I would suggest naming this class a `SessionInterceptor` to keep aligned with `TableIntercetor`. And you can place this class in `interceptor/` package.

Shuo-Jia (approved):

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSessionHookManager.java
levy5307: Yes, I can put it in `interceptor/`, but there would be two interceptorManager, which are `InterceptorManager` and `ReplicaSessionInterceptorManager`

configuration/pegasus_jaas.conf
levy5307: 1.I have tested that this config does't support comment. So I can write a file later to show what`s the meaning of these config item.
2. Yes, our users should have this file in their project.

levy5307: Yes, every our users should have this file in their project

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ClusterManager.java
```diff
@@ -56,7 +58,12 @@ public ClusterManager(ClientOptions opts) throws IllegalArgumentException {
     replicaGroup = getEventLoopGroupInstance(opts.getAsyncWorkers());
     metaGroup = getEventLoopGroupInstance(1);
     tableGroup = getEventLoopGroupInstance(1);
-    enableAuth = opts.isEnableAuth();
+    if (opts.isEnableAuth()) {
+      ReplicaSessionInterceptor hook =
+          new SecurityReplicaSessionInterceptor(
+              opts.getJaasConf(), opts.getServiceName(), opts.getServiceFQDN());
+      ReplicaSessionInterceptorManager.instance().addHook(hook);
```
Shuo-Jia, line 65: Just as your previous commet about `TableInterceptor`: https://github.com/XiaoMi/pegasus-java-client/pull/125#discussion_r481045604, the `ReplicaSessionInterceptor` can be put into `ReplicaSessionInterceptorManager`

configuration/pegasus_jaas.conf
levy5307: I removed the `pegasus_jaas.conf`. So we can add it in later pull request. And I will refactor it in a better way at that time.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
```diff
@@ -39,10 +40,13 @@
   }
 
   public ReplicaSession(
-      rpc_address address, EventLoopGroup rpcGroup, int socketTimeout, boolean enableAuth) {
```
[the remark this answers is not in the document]
levy5307: I don't think we should create a `ReplicaSessionInterceptorManager` for each `ReplicationSession`. Each ClusterManager has a `ReplicaSessionInterceptorManager` is enough

neverchanje (approved):

Shuo-Jia (approved):

PR #136, opened 2020-09-24
feat(security): handle list mechanism response

levy5307:


src/main/java/com/xiaomi/infra/pegasus/rpc/interceptor/SecurityReplicaSessionInterceptor.java
```diff
@@ -41,4 +42,13 @@ public void onConnected(ReplicaSession session) {
...
     negotiation.start();
   }
+
+  public boolean onSendMessage(ReplicaSession session, final ReplicaSession.RequestEntry entry) {
+    // tryPendRequest returns false means that the negotiation is succeed now
+    return isNegotiationRequest(entry) || !session.tryPendRequest(entry);
+  }
+
+  private boolean isNegotiationRequest(final ReplicaSession.RequestEntry entry) {
+    return entry.op.getClass().equals(negotiation_operator.class);
```
Shuo-Jia, line 52: using `instanceof` to judge object type, see https://github.com/XiaoMi/pegasus-java-client/blob/3615cfd978797334cb7beec4b624473642698a63/src/main/java/com/xiaomi/infra/pegasus/rpc/interceptor/CompressionInterceptor.java#L37

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
```diff
@@ -366,6 +370,26 @@ public void run() {
...
+  //   false - pend failed
+  public boolean tryPendRequest(RequestEntry entry) {
+    // double check. the first one doesn't lock the lock.
+    // Because negotiationSucceed only transfered from false to true.
+    // So if it is true now, it will not change in the later.
+    // But if it is false now, maybe it will change soon. So we should use lock to protect it.
+    if (!this.negotiationSucceed) {
+      synchronized (negotiationPendingSend) {
+        if (!this.negotiationSucceed) {
+          negotiationPendingSend.offer(entry);
```
Shuo-Jia, line 384: Why not use `pendingSend` but create new pending object?

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
```diff
@@ -326,6 +326,10 @@ void tryNotifyFailureWithSeqID(int seqID, error_types errno, boolean isTimeoutTa
   }
 
   private void write(final RequestEntry entry, VolatileFields cache) {
+    if (!interceptorManager.onSendMessage(this, entry)) {
+      return;
```
Shuo-Jia, line 330: before `write` seem no need add `interceptor`, you can put `intercepter` into `markSessionConnected` such as https://github.com/XiaoMi/pegasus-java-client/blob/3615cfd978797334cb7beec4b624473642698a63/src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java#L221

It's also means 
```java
isNegotiationRequest(entry) || !session.tryPendRequest(entry);
```
can be put into `onConnected`

src/main/java/com/xiaomi/infra/pegasus/rpc/async/Negotiation.java
```diff
@@ -1,32 +1,38 @@
...
+import java.util.List;
 import javax.security.auth.Subject;
 import javax.security.sasl.Sasl;
 import org.slf4j.Logger;
 
 public class Negotiation {
   private static final Logger logger = org.slf4j.LoggerFactory.getLogger(Negotiation.class);
+  // Because negotiation message is always the first rpc sent to pegasus server,
+  // which will cost much more time. so we set negotiation timeout to 10s here
+  private static final int negotiationTimeoutMS = 10000;
```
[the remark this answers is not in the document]
levy5307: I don't think it must under the user's control. The negotiation process should be not visable to users

src/main/java/com/xiaomi/infra/pegasus/rpc/async/Negotiation.java
```diff
@@ -35,11 +41,13 @@ public void start() {
   }
 
   public void send(negotiation_status status, blob msg) {
-    // TODO: send negotiation message, using RecvHandler to handle the corresponding response.
+    negotiation_request request = new negotiation_request(status, msg);
+    negotiation_operator operator = new negotiation_operator(request);
+    session.asyncSend(operator, new RecvHandler(operator), negotiationTimeoutMS, false);
```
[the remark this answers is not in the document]
levy5307: In IDEA, we can see the attention which shows it is isBackupRequest. So I think there is no need to add `/* isBackupRequest */` here.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
levy5307: We can't use `pendingSend` here.  Because  if the session is connected and the negotiation is not succeed, the session will send the pending messages in `pendingSend`. So if we put it into `pendingSend`, it will procude error.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/Negotiation.java
```diff
@@ -63,19 +72,57 @@ private void handleResponse() throws Exception {
...
-          throw new Exception("Received an unexpected response, status " + resp.status);
+          throw new Exception("unexpected negotiation status: " + resp.status);
+      }
+    }
+  }
+
+  public void onRecvMechanisms(negotiation_response response) throws Exception {
+    checkStatus(response.status, negotiation_status.SASL_LIST_MECHANISMS_RESP);
+
+    String[] matchMechanism = new String[1];
```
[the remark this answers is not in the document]
levy5307: Because the interface of sasl need type of `String[]`. And we can offer more than one mechanism in the later

src/main/java/com/xiaomi/infra/pegasus/rpc/async/Negotiation.java
```diff
@@ -63,19 +72,57 @@ private void handleResponse() throws Exception {
...
+  public void checkStatus(negotiation_status status, negotiation_status expected_status)
+      throws Exception {
+    if (status != expected_status) {
+      throw new Exception("status is " + status + " while expect " + expected_status);
+    }
+  }
+
+  private void negotiationFailed() {
+    status = negotiation_status.SASL_AUTH_FAIL;
+    session.closeSession();
```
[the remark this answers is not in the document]
levy5307: When the `session.closeSession` is called,  `channelInactive` will be called synchronously, which will call `tryNotifyFailureWithSeqID` finallly

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
levy5307: No, we should call this interceptor when a message is sending, which means it will be called twice times if you send two messages. But `onConnected` is called only when the session is connected, and it will only called once.

src/main/java/com/xiaomi/infra/pegasus/rpc/async/Negotiation.java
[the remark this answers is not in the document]
neverchanje: Yes. Only if you use IDE. But it's not for all code viewers.

src/test/java/com/xiaomi/infra/pegasus/rpc/async/NegotiationTest.java
```diff
@@ -18,21 +18,86 @@
...
+  public void testRecvMechanisms() {
+    Negotiation mockNegotiation = Mockito.spy(negotiation);
+    SaslWrapper mockSaslWrapper = Mockito.mock(SaslWrapper.class);
+    mockNegotiation.saslWrapper = mockSaslWrapper;
+
+    Mockito.doNothing().when(mockNegotiation).send(any(), any());
+    try {
+      Mockito.when(mockNegotiation.saslWrapper.init(any())).thenReturn(new byte[0]);
+    } catch (Exception ex) {
+      Assert.fail();
```
Shuo-Jia, line 80: you can use `Assertions.assertThrows` or `Assertions.assertDoesNotThrow` to asset the result replace of `try..catch`

src/test/java/com/xiaomi/infra/pegasus/rpc/async/NegotiationTest.java
```diff
@@ -18,21 +18,87 @@
...
+    }
+
+    status = negotiation_status.SASL_LIST_MECHANISMS_RESP;
+    try {
+      negotiation.checkStatus(status, expectedStatus);
+      Assert.fail();
+    } catch (Exception e) {
+    }
+  }
```
Shuo-Jia, line 69: Suggest here better replace using `assertThrow` and test exception message

Shuo-Jia (dismissed):

Shuo-Jia (approved):

src/main/java/com/xiaomi/infra/pegasus/rpc/async/ReplicaSession.java
```diff
@@ -381,6 +385,26 @@ public void run() {
...
+      synchronized (negotiationPendingSend) {
+        if (!this.negotiationSucceed) {
+          negotiationPendingSend.offer(entry);
+          return true;
+        }
+      }
+    }
+
+    return false;
+  }
```
neverchanje, line 406: 1. Your design here binds the responsibility of negotiation to ReplicaSession again. Why couldn't ReplicaSessionInterceptor hold the pending queue? 

2. negotiationPendingSend is declared final, but you offer items here.

levy5307: It's hard to hold pending queue by `ReplicaSessionInterceptor`. 
1. `ReplicaSessionInterceptor` is a singleton, which will serves all of the ReplicaSession. So if we do this like you said, we must use a coarse-grained lock, which will produce low effiency
2. If we use `ReplicaSessionInterceptor` to do this, we must implement a lot of logic in it. For example timeout handling, which will mantain a timeout pendingResponse queue and a PendingSend, and so on. It's too complicated, and there is a lot of overlap code with ReplicaSession. And we should modify ReplicaSession to support it.

neverchanje (approved):

Shuo-Jia (approved):

PR #137, opened 2020-09-24
feat(security): handle mechanism selected response

levy5307:


src/main/java/com/xiaomi/infra/pegasus/security/SaslWrapper.java
```diff
@@ -51,4 +52,17 @@
               return saslClient.getMechanismName().getBytes(Charset.defaultCharset());
             });
   }
+
+  public blob getInitialResponse() throws PrivilegedActionException {
```
neverchanje, line 56: This is function name is not clear enough. What does that mean?

neverchanje (approved):

src/main/java/com/xiaomi/infra/pegasus/security/SaslWrapper.java
levy5307: I added a comment in front of func getInitialResponse. pls have a look

neverchanje (approved):

Shuo-Jia (approved):

PR #139, opened 2020-09-25
feat(security): implement AuthProtocol and Credential classes

levy5307:


Shuo-Jia (approved):

src/main/java/com/xiaomi/infra/pegasus/rpc/interceptor/SecurityReplicaSessionInterceptor.java
```diff
@@ -32,22 +37,32 @@
 
   private String serviceName;
   private String serviceFqdn;
+
+  // JAAS internal class, Ref:
+  // https://docs.oracle.com/javase/7/docs/technotes/guides/security/jaas/JAASRefGuide.html
+  //
+  // To authorize access to resources, applications first need to authenticate the source of the
+  // request. The JAAS framework defines the term subject to represent the source of a request. A
+  // subject may be any entity, such as a person or a service
```
neverchanje, line 46:
```suggestion
  // Subject is a JAAS internal class, Ref:
  // https://docs.oracle.com/javase/7/docs/technotes/guides/security/jaas/JAASRefGuide.html
  //
  // To authorize access to resources, applications first need to authenticate the source of the
  // request. The JAAS framework defines the term "subject" to represent the source of a request. A
  // subject may be any entity, such as a person or a service.
```

src/main/java/com/xiaomi/infra/pegasus/rpc/interceptor/SecurityReplicaSessionInterceptor.java
```diff
@@ -32,22 +37,32 @@
...
 
     try {
-      loginContext = new LoginContext("client", new TextCallbackHandler());
+      // actual authentication: authenticate the Subject
+      subject = new Subject();
+      loginContext =
+          new LoginContext(
+              "pegasus-client",
+              subject,
+              new TextCallbackHandler(),
```
neverchanje, line 63: I've no found any refs why use `TextCallbackHandler`. Could you explain it in the comment?

src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
```diff
@@ -379,6 +401,28 @@ public Builder serviceFQDN(String serviceFQDN) {
...
+    /**
+     * kerberos principal. Defaults to {@literal ""}, see {@link #DEFAULT_PRINCIPAL}.
+     *
+     * @param principal
+     * @return {@code this}
+     */
+    public Builder principal(String principal) {
+      this.principal = principal;
+      return this;
+    }
```
neverchanje, line 424: I'd recommend you wrap these two options into a Credential class in order to support more types of authentication with less modification. Your current implementation limits the auth method to Kerberos only. The SecurityReplicaSessionInterceptor also assumes that the mechanism is definitely Kerberos. You should take extensibility into account.

Consider this design. A Credential could possibly be keytab/principle, or AK/SK that's provided by the cloud provider. You can extend Credential to a KerberosCredential for Kerberos authentication.

```java
public interface Credential {
  public Mechanism getMechanism();
}

public interface Mechanism {
  void authenticate() throws PException;
}
```

To load from the configuration file, KerberosCredential must have a constructor that accepts Configuration:

```java
class KerberosCredential {
  KerberosCredential(Configuration cfg);

  Mechanism getMechanism() {
    return new JAASMechanism(this);
  }
}

class JAASMechanism {
    void authenticate() throws PException; // use LoginContext
}
```

To determine the auth method we use, an option "auth_mechanism=kerberos" must be given. It can replace "enable_auth" for skipping auth. "auth_mechanism=none" or "auth_mechanism=" mean no auth.

src/main/java/com/xiaomi/infra/pegasus/rpc/interceptor/SecurityReplicaSessionInterceptor.java
```diff
@@ -32,22 +37,34 @@
...
-      throws IllegalArgumentException {
-    this.serviceName = serviceName;
-    this.serviceFqdn = serviceFqdn;
+  public SecurityReplicaSessionInterceptor(ClientOptions options) throws IllegalArgumentException {
+    this.serviceName = options.getServiceName();
+    this.serviceFqdn = options.getServiceFQDN();
 
     try {
-      loginContext = new LoginContext("client", new TextCallbackHandler());
+      // actual authentication: authenticate the Subject
```
neverchanje, line 57: If loginContext.login is the actual authentication, what does Negotiation do? This comment is confusing.

levy5307: To authorize access to resources, applications first need to authenticate the source of the request. The JAAS framework defines the term "subject" to represent the source of a request.

neverchanje: So it's one of the authentication steps. It doesn't represent the entire "actual authentication".


```suggestion
      // authenticate the Subject (the source of the request)
```

src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
```diff
@@ -345,37 +332,24 @@ public Builder metaQueryTimeout(Duration metaQueryTimeout) {
     }
 
     /**
-     * Whether to enable authentication. Defaults to {@literal false}, see {@link
-     * #DEFAULT_ENABLE_AUTH}.
+     * authentiation protocol. Defaults to {@literal ""}, See {@link #DEFAULT_AUTH_PROTOCOL}
```
neverchanje, line 335:
```suggestion
     * The authentiation protocol to use. Available protocols are:
     * <p><ul>
     *  <li>""</li>
     * </ul></p>
     * Defaults to {@literal ""}, See {@link #DEFAULT_AUTH_PROTOCOL}
```

I don't see any available protocols? What does empty string mean?

levy5307: empty means the authentiation is disabled

src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
```diff
@@ -31,6 +32,7 @@
  *          .falconPushInterval(Duration.ofSeconds(10))
  *          .metaQueryTimeout(Duration.ofMillis(5000))
  *          .enableAuth(false)
```
Shuo-Jia, line 34: `enableAuth` deleted

src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
```diff
@@ -58,9 +58,7 @@
   public static final Duration DEFAULT_FALCON_PUSH_INTERVAL = Duration.ofSeconds(10);
   public static final boolean DEFAULT_ENABLE_WRITE_LIMIT = true;
   public static final Duration DEFAULT_META_QUERY_TIMEOUT = Duration.ofMillis(5000);
-  public static final boolean DEFAULT_ENABLE_AUTH = false;
```
Shuo-Jia, line 61: delete `DEFAULT_ENABLE_AUTH`, how judge open or close? `DEFAULT_AUTH_PROTOCOL = ""` means close?

levy5307: yes, you are right

configuration/pegasus.properties
```diff
@@ -5,5 +5,8 @@ enable_perf_counter = false
 perf_counter_tags = cluster=onebox,app=unit_test
 push_counter_interval_secs = 10
 meta_query_timeout = 5000
-service_name = ""
-service_fqdn = ""
+auth_protocol =
+service_name =
+service_fqdn =
+keytab =
+principal =
```
neverchanje, line 12:
```suggestion
kerberos_service_name =
kerberos_service_fqdn =
kerberos_keytab =
kerberos_principal =
```

"kerberos_xxx"  helps users with bare knowledge of authentication quickly understand the meaning of these configs.
I don't even know what "principal" means actually if given no context of authentication.

src/main/java/com/xiaomi/infra/pegasus/security/KerberosProtocol.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Map;
+import javax.security.auth.Subject;
+import javax.security.auth.login.AppConfigurationEntry;
+import javax.security.auth.login.Configuration;
+import javax.security.auth.login.LoginContext;
+import javax.security.auth.login.LoginException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class KerberosProtocol implements AuthProtocol {
```
neverchanje, line 33:
```suggestion
class KerberosProtocol implements AuthProtocol {
```

I think KerberosCredential is user-facing. But KerberosProtocol can be hidden.

src/main/java/com/xiaomi/infra/pegasus/security/KerberosCredential.java
```diff
@@ -0,0 +1,50 @@
...
+  public static final String DEFAULT_SERVICE_FQDN = "";
+  public static final String DEFAULT_KEYTAB = "";
+  public static final String DEFAULT_PRINCIPAL = "";
+
+  private String serviceName;
+  private String serviceFqdn;
+  private String keyTab;
+  private String principal;
+
+  public KerberosCredential(Configuration config) {
```
neverchanje, line 39:
```suggestion
  KerberosCredential(Configuration config) {
```

This constructor is called only by `Credential.create`. It shouldn't be public.

The whole class can be hidden actually. I think you haven't considered that users could create a `Credential` by hard-coded configs, not only read config file. I recommend you add a Builder class for KerberosCredential.

```java
public class KerberosCredentialBuilder {
  public void serviceName(String svcName) {
    credential.serviceName = svcName;
  }
  ...
  public Credential build() {
    return credential;
  }
  private KerberosCredential credential;
}
```

src/main/java/com/xiaomi/infra/pegasus/security/KerberosCredential.java
```diff
@@ -0,0 +1,156 @@
...
+        + principal
+        + '\''
+        + '}';
+  }
+
+  public static class Builder {
+    private String serviceName;
+    private String serviceFqdn;
+    private String keyTab;
+    private String principal;
```
neverchanje, line 101:
```suggestion
  public static class Builder {
    private Credential cred;    
```

This is a repeat-your-code.

```
    /**
     * kerberos service name. Defaults to {@literal ""}, see {@link #DEFAULT_SERVICE_NAME}
     *
     * @param serviceName
     * @return {@code this}
     */
    public Builder serviceName(String serviceName) {
      this.cred.serviceName = serviceName;
      return this;
    }
```

I think you don't have to redeclare those private variables

neverchanje (approved):

Shuo-Jia (approved):

PR #148, opened 2021-01-25
fix: make sure return "maxFetchCount" record if valid record is enough when multiGetSortKeys

Shuo-Jia:


src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1067,17 +1067,62 @@ public MultiGetResult multiGet(
...
+            String.format(
+                "getting pegasusScanner takes too long time when multiGet hashKey=%s, startSortKey=%s, stopSortKey=%s, timeUsed=%s",
+                hashKeyStr, startSortKeyStr, stopSortKeyStr, lastCheckTime - startTime));
+      }
+
+      int remainingTime;
+      Pair<Pair<byte[], byte[]>, byte[]> pairs;
+      while ((pairs = pegasusScanner.next()) != null
+          && multiGetResult.values.size() < maxFetchCount) {
+        multiGetResult.values.add(Pair.of(pairs.getLeft().getValue(), pairs.getValue()));
```
levy5307, line 1104: move this line to L1119

src/test/java/com/xiaomi/infra/pegasus/client/TestBasic.java
```diff
@@ -2629,4 +2629,91 @@ private void testWriteSizeLimit(PegasusClientInterface client) {
...
+    List<Pair<byte[], byte[]>> values = new ArrayList<>();
+    // case 1: scan all record: if persistent record count >= maxFetchCount, it must return
+    // maxFetchCount records
+    boolean case1 =
+        client.multiGet(
+            tableName, hashKey.getBytes(), null, null, new MultiGetOptions(), 5, -1, values);
+    Assert.assertFalse(case1);
+    Assert.assertEquals(5, values.size());
+    Assert.assertEquals("persistent_0", new String(values.get(0).getKey()));
+    Assert.assertEquals("persistent_4", new String(values.get(4).getKey()));
```
levy5307, line 2650: add assert about `allFetched`

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1067,17 +1067,62 @@ public MultiGetResult multiGet(
       int timeout /*ms*/)
       throws PException {
     if (timeout <= 0) timeout = defaultTimeout;
+
+    String hashKeyStr = hashKey == null ? "" : new String(hashKey);
+    String startSortKeyStr = startSortKey == null ? "" : new String(startSortKey);
+    String stopSortKeyStr = stopSortKey == null ? "" : new String(stopSortKey);
```
neverchanje, line 1073: Not all hashkey/sortkey/value use UTF-8 encoding. Are you aware of this?
Printing String from bytes may display unprintable characters, which may not be confusing rather than helpful to users.

I recommend we should simply offer the problem to the caller.

```
new TimeoutException("getting scanner times out")
```

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1067,17 +1067,62 @@ public MultiGetResult multiGet(
...
+      if (pegasusScanner.next() == null) {
+        multiGetResult.allFetched = true;
+      }
+      return multiGetResult;
     } catch (TimeoutException e) {
-      throw PException.timeout(
-          metaList, table.getTableName(), new Request(hashKey, maxFetchCount), timeout, e);
-    } catch (ExecutionException e) {
-      throw new PException(e);
+      throw new PException(new ReplicationException(error_code.error_types.ERR_TIMEOUT));
```
neverchanje, line 1125: Shouldn't we use `PException.timeout()` ? 

https://github.com/XiaoMi/pegasus-java-client/blob/6081ceaefa08014d76aefecd59bf4351624f2c6c/src/main/java/com/xiaomi/infra/pegasus/client/PException.java

levy5307 (approved):

hycdong (approved):

neverchanje (approved):

PR #151, opened 2021-03-16
refactor: use scanner to refactor range operation(get&delete)

Shuo-Jia:


src/main/java/com/xiaomi/infra/pegasus/client/request/range/DeleteRange.java
```diff
@@ -0,0 +1,72 @@
...
+                      ScanOptions scanOptions = new ScanOptions();
+                      scanOptions.noValue = true;
+                      PegasusScannerInterface scanner =
+                          table.getScanner(hashKey, startSortKey, stopSortKey, scanOptions);
+                      List<byte[]> sortKeys = new ArrayList<>();
+                      Pair<Pair<byte[], byte[]>, byte[]> pairs;
+                      while ((pairs = scanner.next()) != null) {
+                        sortKeys.add(pairs.getKey().getValue());
+                        if (sortKeys.size() == scanOptions.batchSize) {
+                          table.multiDel(hashKey, sortKeys, 0);
```
levy5307, line 50:
```
table.multiDel(hashKey, sortKeys, 0);
sortKeys.clear();
```
=====>
```
break;
```

Shuo-Jia: Only `next=null` allow `break`

PR #153, opened 2021-04-01
fix: validate the client config value

Shuo-Jia:
Considering compatibility, the `validate` use `assert` but not `Exception`,  avoid to let user must update code  to  `throw Exption`

### DIFF
--- a/src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
@@ -245,6 +245,7 @@ public class ClientOptions {
     public Builder metaServers(String metaServers) {
       this.metaServers = metaServers;
       return this;
+
     }
 
     /**
--- a/src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
@@ -248,26 +248,28 @@ public class ClientOptions {
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
```diff
@@ -369,6 +373,10 @@ public Builder credential(Credential credential) {
     public ClientOptions build() {
       return new ClientOptions(this);
     }
+
+    private void validatePositive(long value) {
```
neverchanje, line 377:
```suggestion
    private static void validatePositiveNum(long value) {
```

neverchanje (approved):

src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
```diff
@@ -369,6 +373,10 @@ public Builder credential(Credential credential) {
     public ClientOptions build() {
       return new ClientOptions(this);
     }
+
+    private static void validatePositiveNum(long value) {
+      assert value > 0 : String.format("must pass positive value: %d", value);
```
levy5307, line 378: assert value > 0 : String.format("must pass positive value: %ld", value);

levy5307 (approved):

src/main/java/com/xiaomi/infra/pegasus/client/ClientOptions.java
Shuo-Jia: %d can be for int, long, byte in Java

PR #154, opened 2021-04-01
refactor: refactor getRange operation using scanner

Shuo-Jia:


src/main/java/com/xiaomi/infra/pegasus/client/request/range/ScanResult.java
```diff
@@ -0,0 +1,70 @@
...
+  public ScanResult(List<Pair<Pair<byte[], byte[]>, byte[]>> results, boolean allFetched) {
+    this.results = results;
+    this.allFetched = allFetched;
+  }
+
+  public ScanResult(PException exception) {
+    this.exception = exception;
+  }
+
+  public MultiGetResult convertMultiGetResult() {
```
levy5307, line 45: add comment for these two functions

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -1173,16 +1178,13 @@ public int batchMultiGet2(
...
-    sortKeysResult.keys = new ArrayList<>();
-    ScanOptions options = new ScanOptions();
-    options.noValue = true;
-    ScanRangeResult result = scanRange(hashKey, null, null, options, maxFetchCount, timeout);
-    for (Pair<Pair<byte[], byte[]>, byte[]> pair : result.results) {
-      sortKeysResult.keys.add(pair.getLeft().getValue());
-    }
-    sortKeysResult.allFetched = result.allFetched;
-    return sortKeysResult;
+    ScanOptions scanOptions = new ScanOptions();
```
levy5307, line 1181: should we reserver the old interface and add a new interface for these code?

Shuo-Jia: Actually, if like `multiGet`, origin `multiGetSortKeys` should also be retained and only marked `Deprecated`. but the API has been refactored by `scan` in previous PR and some user has use it, here I just keep it.  @neverchanje @hycdong can give some suggestion

src/test/java/com/xiaomi/infra/pegasus/client/TestBasic.java
```diff
@@ -2632,87 +2631,27 @@ private void testWriteSizeLimit(PegasusClientInterface client) {
...
-    Assert.assertFalse(caseD1.allFetched);
-    Assert.assertEquals(5, caseD1.keys.size());
+    MultiGetSortKeysResult result1 = table.multiGetSortKeys(hashKey.getBytes(), 5, -1, 0);
+    Assertions.assertFalse(result1.allFetched);
+    Assertions.assertEquals(5, result1.keys.size());
     for (int i = 0; i <= 4; i++) {
-      Assertions.assertEquals("persistent_" + i, new String(caseD1.keys.get(i)));
+      Assertions.assertEquals("persistent_" + i, new String(result1.keys.get(i)));
     }
     // case D1: maxFetchCount < 0, return all valid record
```
hycdong, line 2649: case A,B,C have been moved to TestRange.java, how about update comment for `case D1`?

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
hycdong: I think it is okay to update code in function multiGetSortKeys, because it is already implemented by scan.

PR #158, opened 2021-06-03
feat(security): refresh tgt in background

levy5307:
If we use kerberos in jaas, we should refresh tgt in background to keep the authentication succeed.
For more details: https://levy5307.github.io/blog/jaas-tgt-renewal/
So in this pull request, I add a background thread to check and refresh the tgt periodically.

### Manual Test

Print the life time of tgt obtain in java client. And we can see the life time of this tgt is updated before it is expired(At line 8).

1. the remain lifetime of tgt: 69 seconds
2. the remain lifetime of tgt: 59 seconds
3. the remain lifetime of tgt: 49 seconds
4. the remain lifetime of tgt: 39 seconds
5. the remain lifetime of tgt: 29 seconds
6. the remain lifetime of tgt: 19 seconds
7. ***the remain lifetime of tgt: 9 seconds***
8. ***the remain lifetime of tgt: 69 seconds***
9. the remain lifetime of tgt: 59 seconds
10. the remain lifetime of tgt: 49 seconds

### DIFF
--- a/src/main/java/com/xiaomi/infra/pegasus/security/KerberosProtocol.java
+++ b/src/main/java/com/xiaomi/infra/pegasus/security/KerberosProtocol.java
@@ -52,7 +52,7 @@ class KerberosProtocol implements AuthProtocol {
   private String serviceFqdn;
   private String keyTab;
   private String principal;
-  final int CHECK_TGT_INTEVAL = 30 * 1000;
+  final int CHECK_TGT_INTEVAL_SECONDS = 10;
 
   KerberosProtocol(String serviceName, String serviceFqdn, String keyTab, String principal)
       throws IllegalArgumentException {
@@ -62,7 +62,7 @@ class KerberosProtocol implements AuthProtocol {
     this.principal = principal;
 
     this.login();
-    scheduleRefreshTGT();
+    scheduleCheckTGTAndRelogin();
     logger.info("login succeed, as user {}", subject.getPrincipals().toString());
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/xiaomi/infra/pegasus/security/KerberosProtocol.java
```diff
@@ -81,6 +77,80 @@ public boolean isAuthRequest(final ReplicaSession.RequestEntry entry) {
     return entry.op instanceof negotiation_operator;
   }
 
+  private void scheduleCheckTGTAndRelogin() {
+    ScheduledExecutorService service = Executors.newSingleThreadScheduledExecutor();
```
Shuo-Jia, line 81: define global final

src/main/java/com/xiaomi/infra/pegasus/security/KerberosProtocol.java
```diff
@@ -41,32 +48,21 @@
...
   // subject may be any entity, such as a person or a service.
   private Subject subject;
-  // The LoginContext class provides the basic methods used to authenticate subjects, and provides a
-  // way to develop an application independent of the underlying authentication technology
-  private LoginContext loginContext;
   private String serviceName;
   private String serviceFqdn;
+  private String keyTab;
+  private String principal;
+  final int CHECK_TGT_INTEVAL_SECONDS = 10;
```
Shuo-Jia, line 55: from 
> 公司内部的这个renew till时间是10年，所以我们平时测试的话，确实不容易看出问题

I understand the refresh is not necessary in some case, but the 10s task will Increase client load, so better add a config to whether enable auto-refresh?

levy5307: The refresh is necessary. Because the renew till time maybe be changed someday. And I think the timer runs every 10 seconds will not cost too much

Shuo-Jia (approved):

hycdong (approved):

PR #175, opened 2022-02-08
feat: add a batch get client interface which uses 'BATCH_GET' rpc for read optimization

cauchy1988:
corresponding server side issue is ： https://github.com/apache/incubator-pegasus/issues/884

 this newly added client interface  firstly split the needed read items into groups according to related partition and then fetch every grouped-data from 'BATCH_GET' rpc


levy5307:
In my opinion, batchGet3 is not good enough. Better to make it more clear.

src/test/java/com/xiaomi/infra/pegasus/client/TestBatchGet3.java
```diff
@@ -0,0 +1,121 @@
...
+      oneItem.sortKey = ("sortedKey_" + i + "_" + System.currentTimeMillis()).getBytes();
+      oneItem.value = ("value_" + i + "_" + System.currentTimeMillis()).getBytes();
+      pairs.add(Pair.of(oneItem.hashKey, oneItem.sortKey));
+      values.add(oneItem.value);
+      batchSetItems.add(oneItem);
+    }
+
+    List<PException> result = new ArrayList<>();
+    table.batchSet2(batchSetItems, result, 0);
+    for (int i = 0; i < 101; ++i) {
```
levy5307, line 87: It's better to define a const var for `101`

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -985,6 +1036,94 @@ public void batchGet(List<Pair<byte[], byte[]>> keys, List<byte[]> values, int t
...
+        futures.add(null);
+        continue;
+      }
+      emptyResults.add(false);
+      futures.add(asyncBatchGet(request, timeout));
+    }
+
+    List<Map<Pair<String, String>, byte[]>> resultMapList = new ArrayList<>();
+    for (int i = 0; i < this.table.getPartitionCount(); i++) {
+      if (emptyResults.get(i)) {
```
Shuo-Jia, line 1092: `emptyResults` is redundant, you use `requestList.get(i).keys.isEmpty` is ok

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -985,6 +1036,94 @@ public void batchGet(List<Pair<byte[], byte[]>> keys, List<byte[]> values, int t
...
+      int index = responseIndex.get(i);
+      if (null != partitionToPException.get(index)) {
+        results.add(Pair.of(partitionToPException.get(index), null));
+        continue;
+      }
+
+      byte[] value =
+          resultMapList
+              .get(index)
+              .get(Pair.of(new String(keys.get(i).getLeft()), new String(keys.get(i).getRight())));
```
Shuo-Jia, line 1119: `Pair.of` will return new object, though it's value is same. Do you make sure `map` can get value by the key?

cauchy1988: why not the map cannot get the value by the key ? is there any pitfall?  my  unit test pass, it looks ok;

src/main/java/com/xiaomi/infra/pegasus/client/PegasusClientInterface.java
```diff
@@ -149,6 +149,10 @@ public int batchGet2(
       String tableName, List<Pair<byte[], byte[]>> keys, List<Pair<PException, byte[]>> results)
       throws PException;
 
+  public int batchGet3(
```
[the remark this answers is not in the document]
cauchy1988: ok

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
```diff
@@ -399,6 +399,57 @@ public void operationComplete(Future<MultiGetResult> future) throws Exception {
...
+              handleReplicaException(
+                  new Request(request.keys.get(0).hash_key.data), promise, op, table, timeout);
+            } else if (gop.get_response().error != 0 && gop.get_response().error != 7) {
+              // rocksdb::Status::kOk && rocksdb::Status::kIncomplete
+              promise.setFailure(new PException("rocksdb error: " + gop.get_response().error));
+            } else {
+              BatchGetResult result = new BatchGetResult();
+              result.allFetched = (gop.get_response().error == 0);
+              result.valueMap = new HashMap<>();
+              for (full_data oneData : gop.get_response().data) {
```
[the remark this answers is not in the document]
cauchy1988: ok

Shuo-Jia:
> add a batch get client interface which firstly split the needed read items into groups according to related partition and then fetch every grouped-data from 'BATCH_GET' rpc

title is too lang, the detail  that you should add into `description`

Shuo-Jia:
> In my opinion, the name of batchGet3 is not good enough. Better to make it more clear.

Yeah, but It's a little hard to name. `groupGet` or `batchGetGroup`..... this maybe still not good

cauchy1988:
> > In my opinion, the name of batchGet3 is not good enough. Better to make it more clear.
> 
> Yeah, but It's a little hard to name. `groupGet` or `batchGetGroup`..... this maybe still not good

what about batchGetByPartition ？

cauchy1988:
> > add a batch get client interface which firstly split the needed read items into groups according to related partition and then fetch every grouped-data from 'BATCH_GET' rpc
> 
> title is too lang, the detail that you should add into `description`

ok

src/main/java/com/xiaomi/infra/pegasus/client/PegasusTable.java
Shuo-Jia: OK, I see,  `Pair` overload `hashCode()` and `equal` method

acelyc111:
How about batchGetByPartitions, since this request is on multiple partions.

idl/rrdb.thrift
```diff
@@ -147,6 +147,29 @@ struct multi_get_response
     6:string        server;
 }
 
+struct batch_get_request {
```
acelyc111, line 150: After we reach an agreement to the API name, I think it's better to keep these thrift structure names to be the same/similar to the API's name too. Of course, the server side should be updated too.

cauchy1988: I think api name need not like inner rpc request struct name since the batchGet2 api interface used very different named innser rpc request struct;   in my opinion api name has nothing to do with the inner rpc equest name ; 
  'batch_get_request' is more simpler and can be easily differentiated from "single get  request" and "multi get request"; that is enough~

acelyc111 (approved):

